PR #117, opened 2015-03-06
Change Rack::Request URL in the README

lucaspinto:
Now that rubyforge is dead, we should replace its URL for the GitHub one.


ktheory:
Nice catch. Thanks.

I prefer linking to the docs instead of code, so I tweaked this in 77d0493.

lucaspinto:
No problem!

Thank you for your work!

Issue #113, opened 2015-02-05
Ability to reset specific throttles


rwdaigle:
I am finding it very difficult to do the following, which makes me think I'm using this software incorrectly or for the wrong use-case:

I am currently throttling login attempts to my app. When a user resets their password, I would like to reset the throttle for that specific user. It seems as if the Rack::Attack API is very unfriendly to such under-the-cover shenanigans. There is no throttle-specific reset function, nor is it particularly easy to recreate the key used to store the incrementer to allow for manual resetting.

What is the right way to go about this? Thanks!

ktheory:
> I am currently throttling login attempts to my app.

Excellent! That's a good security practice.

> When a user resets their password, I would like to reset the throttle for that specific user.

Can you explain further the use case?

Generally, this doesn't come up b/c login throttles have such a high threshold that legit attempts (e.g. users who forgot their password) don't reach the threshold. It just catches malicious crackers. So it's moot to reset the throttle when the password changes.

rwdaigle:
Hi @ktheory,

> Generally, this doesn't come up b/c login throttles have such a high threshold

In our environment, and subject to the rigors of PCI compliance, our threshold limit is definitely within reach for some legitimate login attempts.

Also, it might be nice to have such functionality exposed, or at least a primitive available, to aid in resetting test environments. I currently have to do the following to un-block my test login attempts, which seems to be a bit heavy-handed:

```ruby
Rack::Attack.cache.store.clear
```

Seems like for both situations, having a more accessible API would be advantageous?

zmillman:
From the PCI compliance guidelines (copied from http://security.stackexchange.com/a/523)

* 8.5.13 (Limit repeated access attempts by locking out the user ID after not more than six attempts)
* 8.5.14 (Set the lockout duration to thirty minutes or until administrator enables the user ID).

These restrictions only apply to accounts which are accessing cardholder data which is usually only internally authorized employees...but it looks like @rwdaigle is working for Spreedly so there's compliance issues that most other developers wouldn't have.

It looks like it's possible to infer the key name which is built here: https://github.com/kickstarter/rack-attack/blob/master/lib/rack/attack/throttle.rb#L26 and prefixed with the epoch here: https://github.com/kickstarter/rack-attack/blob/master/lib/rack/attack/cache.rb#L17

If you have your throttle configured as
```ruby
Rack::Attack.throttle('logins/email', :limit => 6, :period => 30.minutes) do |req|
[...]

PR #118, opened 2015-03-15
Pass `raw: true` to prevent Ruby de/serialization. This is to make it po...

stanhu:
...ssible

to implement something like:

```store.write(key, 0, :expires_in => expires_in)```

See #113

### DIFF
--- a/lib/rack/attack/store_proxy/redis_store_proxy.rb
+++ b/lib/rack/attack/store_proxy/redis_store_proxy.rb
@@ -13,15 +13,15 @@ module Rack
         end
 
         def read(key)
-          self.get(key)
+          self.get(key, raw: true)
         rescue Redis::BaseError
         end
 
         def write(key, value, options={})
           if (expires_in = options[:expires_in])
-            self.setex(key, expires_in, value)
+            self.setex(key, expires_in, value, raw: true)
           else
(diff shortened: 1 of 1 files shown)


stanhu:
Is it better to pass options to both `read` and `write`? I noticed the Dalli proxy code forced usage of raw mode.

ktheory:
@stanhu: looks good. thanks.

ktheory:
@stanhu: Is this going to cause errors when deployed?

E.g., an app has a bunch of marshalled values in redis already. When they deploy this update, Rack::Attack will read them as raw values instead.

stanhu:
@ktheory, yes it is a potential issue. We could make it default to marshalled values and use the options to override this behavior. But there is an inconsistency already in that the `increment` function uses raw values and the `read` uses marshalled values, which is what caused me hours of debugging. Thoughts?

ktheory:
@stanhu: ok, so the only time Rack::Attack `read`s keys is in Fail/Allow2Ban. And anyone using Fail/Allow2Ban with redis would already be getting these serialization errors.

So it seems like this is strictly an improvement; not a gotcha for any redis users. :smile:

stanhu:
@ktheory: Right, thanks!

PR #119, opened 2015-03-15
Support delete method for Redis

stanhu:
Delete the key outright to support #113


ktheory:
Seems fine. @zmillman: what do you think?

zmillman:
Looks good to me :+1: 

It was silly that Rack::Attack has `delete` for Dalli, but not for Redis

ktheory:
@stanhu mind giving commit 3e0fcc2 a try, and I'll cut a release in the next few days?

PR #126, opened 2015-04-30
Added ability to disable.

hooverlunch:
I wanted the ability to turn off throttling in most of my specs, but turn it on for specs testing throttling itself.

So my test.rb now looks like this:

```
  # Use a lower ODK Collect request limit for tests
  configatron.direct_auth_request_limit = 5

  # Enable rack-attack middleware for protecting against brute-force login attempts,
  # but disable it until needed.
  config.middleware.use Rack::Attack
  Rack::Attack.disable!
```

And then I use `before` and `after` blocks to enable/re-disable it when needed.

Would be great to get this in the main build.

Thanks!

### DIFF
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -76,6 +76,17 @@ class Rack::Attack
       @whitelists, @blacklists, @throttles, @tracks = {}, {}, {}, {}
     end
 
+    def disable!
+      @disabled = true
+    end
+
+    def enable!
+      @disabled = false
+    end
+
+    def disabled?
+      !!@disabled
+    end
   end
 
   # Set defaults
@@ -91,6 +102,8 @@ class Rack::Attack
   end
 
   def call(env)
+    return @app.call(env) if disabled?
+
     req = Rack::Attack::Request.new(env)
 
     if whitelisted?(req)
@@ -109,5 +122,6 @@ class Rack::Attack
   def_delegators self, :whitelisted?,
                        :blacklisted?,
                        :throttled?,
-                       :tracked?
+                       :tracked?,
+                       :disabled?
 end


tiegz:
:+1:

ktheory:
@hooverlunch: did you see the [`Rack::Attack.clear!`](https://github.com/kickstarter/rack-attack/blob/919976cb640c79163796b1e9290165e58f2c7748/lib/rack/attack.rb#L75-L77) method?

You can call that in your `before` block, and restore your rack attack config by `load`ing the config file in your `after` block.

Does that solve your case?

hooverlunch:
Interesting. Seems like a bummer to have to reload the config file. Is
there a reason not to have a disable function?

On 14 May 2015 at 13:56, Aaron Suggs <notifications@github.com> wrote:

> @hooverlunch <https://github.com/hooverlunch>: did you see the
> Rack::Attack.clear!
> <https://github.com/kickstarter/rack-attack/blob/919976cb640c79163796b1e9290165e58f2c7748/lib/rack/attack.rb#L75-L77>
> method?
>
> You can call that in your before block, and restore your rack attack
> config by loading the config file in your after block.
>
> Does that solve your case?
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/kickstarter/rack-attack/pull/126#issuecomment-102117477>
> .
>



-- 
Tom Smyth

Worker-Owner, Sassafras Tech Collective
Specializing in innovative, usable tech for social change
sassafras.coop *·* @sassafrastech

Resident, Touchstone Cohousing
touchstonecohousing.org

ktheory:
First off, thanks for making this Pull Request to make Rack::Attack more useful for you and the community. :smile: 

> Is there a reason not to have a disable function?

My thinking is to optimize for production use. Every production request will check `disabled?` just to handle the less common case that we're in a test w/ Rack::Attack disabled. It's better to do extra work in the test harness than production.

Another idea is to do something like:

```ruby
# Completely untested ;-)
before { Rails.application.middleware.delete Rack::Attack }
after { Rails.application.middleware.use Rack::Attack }
```

I see the value in being able to run integration tests with or without Rack::Attack, and I'd be grateful if you shared the best practice you discover. I'd welcome you sharing whatever best practice  you find in the README or wiki.

hooverlunch:
I tried using middleware delete but it won't let me change middleware
mid-test run.

How about if I changed it so disable! did like clear! but didn't wipe out
the config so that enable! could reload it?



On 14 May 2015 at 15:15, Aaron Suggs <notifications@github.com> wrote:

> First off, thanks for making this Pull Request to make Rack::Attack more
> useful for you and the community. [image: :smile:]
>
> Is there a reason not to have a disable function?
>
> My thinking is to optimize for production use. Every production request
> will check disabled? just to handle the less common case that we're in a
> test w/ Rack::Attack disabled. It's better to do extra work in the test
> harness than production.
>
> Another idea is to do something like:
>
> # Completely untested ;-)
> before { Rails.application.middleware.delete Rack::Attack }
> after { Rails.application.middleware.use Rack::Attack }
>
> I see the value in being able to run integration tests with or without
> Rack::Attack, and I'd be grateful if you shared the best practice you
> discover. I'd welcome you sharing whatever best practice you find in the
> README or wiki.
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/kickstarter/rack-attack/pull/126#issuecomment-102141257>
> .
>



-- 
Tom Smyth

Worker-Owner, Sassafras Tech Collective
Specializing in innovative, usable tech for social change
sassafras.coop *·* @sassafrastech

Resident, Touchstone Cohousing
touchstonecohousing.org

ktheory:
> I tried using middleware delete but it won't let me change middleware
mid-test run.

D'oh!

> How about if I changed it so disable! did like clear! but didn't wipe out
the config so that enable! could reload it?

:+1: Sounds like a good approach.

ktheory:
Closing this for now pending your update. Either re-open this, or open a new PR when you're ready.

Thanks!

PR #128, opened 2015-05-11
Provide the ability to reset Fail2Ban count and ban flag

### LINKED ISSUE #113: Ability to reset specific throttles

stanhu:
This code is in production as a monkey-patch in GitLab v7.10, and I thought it would be helpful for others.

Closes #113

### DIFF
--- a/lib/rack/attack/cache.rb
+++ b/lib/rack/attack/cache.rb
@@ -15,10 +15,7 @@ module Rack
       end
 
       def count(unprefixed_key, period)
-        epoch_time = Time.now.to_i
-        # Add 1 to expires_in to avoid timing error: http://git.io/i1PHXA
-        expires_in = period - (epoch_time % period) + 1
-        key = "#{prefix}:#{(epoch_time/period).to_i}:#{unprefixed_key}"
+        key, expires_in = key_and_expiry(unprefixed_key, period)
         do_count(key, expires_in)
       end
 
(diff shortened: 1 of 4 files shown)


stanhu:
@ktheory Thoughts?

lib/rack/attack/cache.rb
```diff
@@ -30,6 +30,18 @@ def write(unprefixed_key, value, expires_in)
         store.write("#{prefix}:#{unprefixed_key}", value, :expires_in => expires_in)
       end
 
+      def reset_count(unprefixed_key, period)
+        epoch_time = Time.now.to_i
+        # Add 1 to expires_in to avoid timing error: http://git.io/i1PHXA
+        expires_in = period - (epoch_time % period) + 1
+        key = "#{(epoch_time / period).to_i}:#{unprefixed_key}"
```
ktheory: @stanhu this nearly a dupe of lines 18-21. What do you think about extracting it to a private method?

ktheory:
@stanhu: thanks for the nudge. :smile: 

This feature looks useful & I'd like to merge it in.

Could you please add some tests for the `Cache#reset_count` and `Cache#delete` methods? I want to make sure those work across the [various cache stores](https://github.com/kickstarter/rack-attack/blob/919976cb640c79163796b1e9290165e58f2c7748/spec/integration/rack_attack_cache_spec.rb#L19-L26).

stanhu:
@ktheory: Thanks, done. There was already a test for `delete`, although this monkey patch seems necessary for the test only?

https://github.com/kickstarter/rack-attack/blob/919976cb640c79163796b1e9290165e58f2c7748/spec/integration/rack_attack_cache_spec.rb#L4

ktheory:
@stanhu Awesome! Thanks so much. I aim to release version `4.3.0` with this change this afternoon.

ktheory:
@stanhu Just pushed v4.3.0. Thanks for all your improvements recently. You're list release's MVP. :smile:

stanhu:
@ktheory Thank you for being so responsive and creating/maintaining Rack Attack! I'm very happy I can remove the monkey patches now. :)

Issue #29, opened 2013-08-14
could add throttle status to response headers when not exceed throttle


iamroody:
I know we could add these headers to response in controller from "request.env['rack.attack.throttle_data'][name]"， but if we could config add this custom header in rack-attack, it would be great.

PR #138, opened 2015-08-03
Record hit names

zlx:
In #29, @ktheory mentioned: 
> modifying app response headers feels like a big increase in the scope of rack-attack

I agree, but I think the hit names, which means which throttle names have hited is usefully for rate limit implementation.

Hope it can be included soon eagerly.

### DIFF
--- a/lib/rack/attack/throttle.rb
+++ b/lib/rack/attack/throttle.rb
@@ -21,6 +21,8 @@ module Rack
         discriminator = block[req]
         return false unless discriminator
 
+        (req.env['rack.attack.hit_names'] ||= []) << name
+
         current_period = period.respond_to?(:call) ? period.call(req) : period
         current_limit  = limit.respond_to?(:call) ? limit.call(req) : limit
         key            = "#{name}:#{discriminator}"


ktheory:
@zlx: Neat idea, and thanks for making a pull request. :smile: 

One caveat with this implementation:
Rack::Attack stops checking throttles once the first throttle is exceeded ([this line](https://github.com/kickstarter/rack-attack/blob/b631398957b7d2a4fc7da5aa1ab870099fbe0d9b/lib/rack/attack.rb#L56)).

So if you had 3 throttles that all matched a request, and the 2nd throttle was exceeded, `hit_names` would only include the first 2 throttles.

Could explain further how this would be useful for rate limiting?

zlx:
@ktheory  Thank you for reply so quickly.
when I implement rate limit with rack attack, I face the biggest problem is I do not know which throttle is hit, because when I config many throttle and must find the nearest limit and insert it into the response headers.

Base on the use case, if some throttle is exceeded, the request is blocked, so I do not need hit names set headers now.

Maybe you have better idea to record hit names when a throttle is exceeded, let's keep discuss.

ktheory:
> the biggest problem is I do not know which throttle is hit

Good news! :tada: 

That info is already set in the rack environment as `env['rack.attack.matched']`. See [here](https://github.com/kickstarter/rack-attack/blob/b631398957b7d2a4fc7da5aa1ab870099fbe0d9b/lib/rack/attack/throttle.rb#L37-L43) for other metadata about the throttle.

Closing this out since it looks like that addresses your issue. Feel free to reopen if you see fit.

zlx:
@ktheory sorry, I say `which throttle is hit` is mean which throttles are match the request, aka. which throttle will do count for the request.
for example:
I have config three throttle: A, B and C, and if throttle A and throttle B is match current request, so they will do count and I need find the nearest limit from A and B.

`env['rack.attack.matched']` seem record the exceed the limit throttle, and it's just only one(and it must be)

ktheory:
@zlx oh whoops, thanks for clarifying.

Check out `env['rack.attack.throttle_data'].keys` ([these lines of code](https://github.com/kickstarter/rack-attack/blob/master/lib/rack/attack/throttle.rb#L29-L34)). That should give you the throttles that hit.

zlx:
@ktheory Wooow!, Thank you for remind

PR #144, opened 2015-10-23
Updating the readme

jamesprior:
- Notes that the rails cache is also used for fail2ban
- Notes that all fail2ban filters use the same cache for counting and banning
- Expands sample fail2ban filter example to include more matches

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 Rack::Attack is a rack middleware to protect your web app from bad clients.
 It allows *whitelisting*, *blacklisting*, *throttling*, and *tracking* based on arbitrary properties of the request.
 
-Throttle state is stored in a configurable cache (e.g. `Rails.cache`), presumably backed by memcached or redis ([at least gem v3.0.0](https://rubygems.org/gems/redis)).
+Throttle and fail2ban state is stored in a configurable cache (e.g. `Rails.cache`), presumably backed by memcached or redis ([at least gem v3.0.0](https://rubygems.org/gems/redis)).
 
 See the [Backing & Hacking blog post](http://www.kickstarter.com/backing-and-hacking/rack-attack-protection-from-abusive-clients) introducing Rack::Attack.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -128,17 +128,21 @@ end
 `Fail2Ban.filter` can be used within a blacklist to block all requests from misbehaving clients.
 This pattern is inspired by [fail2ban](http://www.fail2ban.org/wiki/index.php/Main_Page).
 See the [fail2ban documentation](http://www.fail2ban.org/wiki/index.php/MANUAL_0_8#Jail_Options) for more details on
-how the parameters work.
+how the parameters work.  All fail2ban filters share the same match counter, using one filter per app is recommended.
```
ktheory: @jamesprior Can you clarify what you mean by `All fail2ban filters share the same match counter`?

jamesprior: Yes!  In testing I tried adding a `Rack::Attack.blacklist('fail2ban wp')` block and a `Rack::Attack.blacklist('fail2ban passwd')` block, each one using a fail2ban filter but checking different criteria.  The cache keys written for both looked like `rack::attack:2409775:fail2ban:count:127.0.0.1` and matching a filter in the first block would also increment the counter for the second block.  Eventually when one of them was triggered it would add a ban rule to the cache for the IP address and it was logged as the first matching blacklist.  What I got from that (and a skimming the fail2ban filter code) was that the fail2ban filters do not include the rack-attack blacklist name in any cache keys and instead they all use the same cache keys for counts and bans.  Changing that would make a fine PR, but trying to explain how it works now in the docs was quick.

ktheory: Yup, you're right. Legit bug. :smile:

I think the recommendation can be to make the discriminator unique to the filter. E.g.:

```ruby
Fail2Ban.filter("passwd-#{req.ip}", ...)
...
Fail2ban.filter("xss-#{req.ip}" ...)
```

ktheory:
Looks good. Thanks, @jamesprior. :smile:

PR #147, opened 2015-11-08
Test ruby 2.2.3

pisaacs:


ktheory:
Good idea, thanks! :sunglasses:

Issue #132, opened 2015-06-26
Multiple blacklist/Fail2Ban not differentiated


mtparet:
If we declare the following rules:
```ruby
Rack::Attack.blacklist('allow2ban login scrapers') do |req|
  Rack::Attack::Allow2Ban.filter(req.ip, :maxretry => 20, :findtime => 1.minute, :bantime => 1.hour) do
    req.path == '/login' and req.post?
  end
end

Rack::Attack.blacklist('oauth scrapers') do |req|
  Rack::Attack::Allow2Ban.filter(req.ip, :maxretry => 20, :findtime => 1.minute, :bantime => 1.hour) do
    req.path == '/oauth/token' and req.post?
  end
end
```

These two different rules will increment the same key although we could think it will increment different keys as we declare different blacklists. (So it will block requests that it should not)

The current solution without modifying the code seems to add an arbitrary name to the discriminator like:
```ruby
Rack::Attack.blacklist('allow2ban login scrapers') do |req|
  Rack::Attack::Allow2Ban.filter("login:#{req.ip}", :maxretry => 20, :findtime => 1.minute, :bantime => 1.hour) do
    req.path == '/login' and req.post?
  end
end

Rack::Attack.blacklist('oauth scrapers') do |req|
  Rack::Attack::Allow2Ban.filter("oauth:#{req.ip}", :maxretry => 20, :findtime => 1.minute, :bantime => 1.hour) do
    req.path == '/oauth/token' and req.post?
  end
end
```

Did I miss something or is it the way to go ? 

(I didn't see anything about this case in the Readme)

ktheory:
Hi @mtparet,

Yes, this is working as intended. You need to give your filters unique names.

mtparet:
Thank you for the confirmation @ktheory!
(I opened the issur for this and to let know others wondering the same thing)

sharq1:
I would assume that the `blacklist`'s name (`blocklist`'s name) is for differentiation. Lost a lot of time until I found explanation here. @grzuy could this be at least documented?

PR #149, opened 2015-11-09
Add note to README.md about scoping of Fail2Ban filters

watsonbox:
I see it has been closed, but I also got tripped up by #132.  I'd like to add a note to the README to avoid confusion.


ktheory:
@watsonbox Good call. :+1: Thank you for improving the docs. :bow:

PR #148, opened 2015-11-08
Test ruby 2.1.7

pisaacs:
Forgot to also update for previous stable, now at 2.1.7 (should have been bundled with #147)

PR #150, opened 2015-11-20
Remove support for ruby 1.9

ktheory:
It's [life ended this past February](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/).

It's not worth testing or supporting.


tiegz:
:metal:

pisaacs:
:+1:  do I have to close and create new PR for #148?

PR #158, opened 2016-01-08
Fix typo in README.md

fighterleslie:
`Rack::Attack::Request` is a subclass of `Rack::Request`, not a subclass of `Rack::Attack`.


ktheory:
:stuck_out_tongue_closed_eyes: Nice catch. Thank you, @fighterleslie.

PR #159, opened 2016-01-12
Add MongoDb as a optional cache store

tmster:
Hi,

Unfortunately in my project I've stopped using either redis or memcached. Because of that, I created optional cache store to support mongodb using mongo-ruby-driver.


ktheory:
Hi @tmster, thanks for using rack-attack an taking the time to implement the mongodb store.

I think it would be much better as a separate plugin rather than bundled with the main gem.

I don't have any MongoDB experience, and reluctant to commit to supporting it. (In contrast, I use memcached and redis frequently).

Thanks again!

PR #170, opened 2016-02-29
Added method banned_until to know the date the user is banned.

mtparet:
### Features
Added method banned_until to know the date the user is banned.


ktheory:
@mtparet: thanks for suggesting this.

Can you please clarify how this would be used?

Do you think you can make it work with memcached (which doesn't have a way to check TTL)?

Lastly, to get this merged, please rebase with a more useful commit message.

I suspect this'll be better implemented as a rack-attack extension rather than in the core library, so I'm closing for now. Feel free to re-open with an expanded description, and working tests.

PR #173, opened 2016-03-25
Add a lazily-loaded, memoized Rack::Attack#ip_addr helper.

tiegz:
[`IPAddr`](http://ruby-doc.org/stdlib-2.3.0/libdoc/ipaddr/rdoc/IPAddr.html) tends to be very useful in tandem with Rack::Attack (checking ipv4 vs ipv6, matching against CIDR blocks, etc). This lazily wraps and memoizes an instance of `IPAddr` as a `Rack::Attack::Request` helper.

cc @ktheory

### DIFF
--- a/lib/rack/attack/request.rb
+++ b/lib/rack/attack/request.rb
@@ -1,3 +1,5 @@
+require 'ipaddr'
+
 # Rack::Attack::Request is the same as ::Rack::Request by default.
 #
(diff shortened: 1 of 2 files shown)


tiegz:
After discussing with @ktheory , closing this in favor of people monkeypatching in apps. Reasoning: not everyone will need IPAddr, and it's easy to patch.

PR #181, opened 2016-07-01
Controversial Language Change Suggestion

renee-travisci:
Hello! 
  At Travis CI we've decided to change our usage of Whitelist/Blacklist to the less racially charged language of Safelist/Blocklist. We are using rack-attack and would love to use an updated version with our language choices if you are open to this change. I'm not sure how you would like to change the versioning of this so I didn't add any changes there. I know this would require users to update all of their whitelist/blacklist method calls to the new verbiage, which might annoy a number of end users. However, we did this with a simple find and replace (and method aliases until this change is welcomed). I added the depreciation warnings for the old method names to help users, however if anyone is relying on the rack.attack.match_type I changed it to safelist/blocklist (wasn't sure how to deprecate that nicely). 
Thanks!
**Renée

PR #183, opened 2016-07-05
More safelist/blocklist refactoring

ktheory:
- Add Rack::Attack namespace to deprecation warning.
- Add deprecated Rack::Attack.blacklisted_response attr methods.
- Drop support for end-of-life ActiveSupport/Dalli versions


ktheory:
Builds on #181

PR #187, opened 2016-07-11
add a spec to specify the behavior of non-matching throttle blocks

mferrier:
It's implied but not obvious that returning falsy from a `throttle` block means that the request does not match and will be disregarded. Added a spec to illustrate this explicitly.

### DIFF
--- a/spec/rack_attack_throttle_spec.rb
+++ b/spec/rack_attack_throttle_spec.rb
@@ -85,3 +85,25 @@ describe 'Rack::Attack.throttle with period as proc' do
     end
   end
 end
+
+describe 'Rack::Attack.throttle with block retuning nil' do
+  before do
+    @period = 60
+    Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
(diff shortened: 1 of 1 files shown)


ktheory:
@mferrier: Thx for the PR.

Please rebase master, it should fix these error in the tests. Good call that these are missing tests.

But it sounds like it was hard to find the info you wanted in the docs. In the [Usage section of the README](https://github.com/kickstarter/rack-attack/blob/88e3ed0efda1e4a9bb04a07ccc943a5efe76362b/README.md#usage):

> Define safelists, blocklists, throttles, and tracks as blocks that return truthy values if matched, falsy otherwise.

I'd welcome a PR making the behavior more obvious.

mferrier:
@ktheory rebased.

You're right, the behaviour *is* documented sufficiently, I just missed the fact that the throttle block return value serves both as match indication (with truthy/falsiness) as well as the cache key in the case of truthiness.

ktheory:
@mferrier thank you! 👍

PR #191, opened 2016-07-18
Adding headers for throttled responses

twiduch:
For all throttled responses added the following headers:

X-RateLimit-Limit
X-RateLimit-Remaining
X-RateLimit-Reset

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,13 @@ For responses that did not exceed a throttle limit, Rack::Attack annotates the e
 request.env['rack.attack.throttle_data'][name] # => { :count => n, :period => p, :limit => l }
 ```
 
+Special headers are added to all throttled responses:
+
+```ruby
+  X-RateLimit-Limit       - number of total allowed requests in a given :period
+  X-RateLimit-Remaining   - number of requests left until end of :period
+  X-RateLimit-Reset       - time remaining to clear counters
+```
 ## Logging & Instrumentation
 
 Rack::Attack uses the [ActiveSupport::Notifications](http://api.rubyonrails.org/classes/ActiveSupport/Notifications.html) API if available.
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -136,18 +136,47 @@ class Rack::Attack
     env['PATH_INFO'] = PathNormalizer.normalize_path(env['PATH_INFO'])
     req = Rack::Attack::Request.new(env)
 
-    if safelisted?(req)
-      @app.call(env)
-    elsif blocklisted?(req)
-      self.class.blocklisted_response.call(env)
-    elsif throttled?(req)
-      self.class.throttled_response.call(env)
-    else
-      tracked?(req)
-      @app.call(env)
+    response = case
+      when safelisted?(req) then @app.call(env)
+      when blocklisted?(req) then self.class.blocklisted_response.call(env)
+      when throttled?(req) then self.class.throttled_response.call(env)
+      else 
+        tracked?(req)
+        @app.call(env)
     end
+    add_throttle_headers(response, env)
   end
 
+  # always returns response object
+  def add_throttle_headers(response, env)
+    return response unless env['rack.attack.throttle_data']
+    data = env['rack.attack.match_data'] || env['rack.attack.throttle_data']
+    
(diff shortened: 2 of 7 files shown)


ktheory:
@twiduch Thanks for helping make rack-attack better!

(For reference, this was discussed in #29.)

I'd rather not change the default responses, since whether to publicly expose this throttle rate depends on the use-case. Developers can always opt-in by customizing their responses.

This also introduces some logic to handle `X-RateLimit-Reset`. Could you explain why that's important for your use case?

twiduch:
@ktheory I just believe this should be the default option. Specially when different requests have different limits - e.g. POST 2/10sec, GET 20/15sec
That allows consumer to queue the calls appropriately.

If you do not want to set headers, I understand. You just prefer to move the logic to devs. However, they do not have info about expiry
The information delivered currently:
  "rack.attack.throttle_data"=>{"api/get/ip"=>{:count=>11, :period=>60, :limit=>20}}
should be
  "rack.attack.throttle_data"=>{"api/get/ip"=>{:count=>11, :period=>60, :limit=>20, :expires_in=>7}}
Without that you cannot set 'X-RateLimit-Reset'

In my opinion most APIs need that info. We could make it optional, if you think that flexibility is needed. 
I mean: by default set headers automatically, but with the ability to switch that off in a configuration file.

class Rack::Attack
  self.throttle_headers = false
end

twiduch:
One more thing. 
When throttle is hit, with that [line](https://github.com/kickstarter/rack-attack/blob/f5f08d56e5869d9b9505f2625eba9b8641f11554/lib/rack/attack.rb#L144) of code - no info is available for devs, and default header ['Retry-After' ](https://github.com/kickstarter/rack-attack/blob/f5f08d56e5869d9b9505f2625eba9b8641f11554/lib/rack/attack.rb#L128) is set

wpolicarpo:
Definitely, would be great to have this as an opt-in option.

@ktheory even if you don't agree with the idea of exposing all these headers, at least we could find a way of exposing the expire window in case we want to add `X-RateLimit-Reset` header.

ktheory:
I understanding what you're asking for, but what's needed here is a use-case explaining why `X-RateLimit-Reset` is important.

@twiduch:
> In my opinion most APIs need that info

What's an example? What's your use-case?

wpolicarpo:
@ktheory in my case, I have to give the consumers the information on when their window will reset, so that they can balance the requests based on that.

Basically, they need to calculate how many requests they can make before the next window and if there are few, they can choose between the most important ones.

twiduch:
@ktheory 
My particular use-case is that I allow 1 request for 10/sec and 5 requests / 100sec. That is normally more than enough. But if a client needs more, he needs information when to send the next call. That is good for them and my server.

That said, I think a specific use-case is irrelevant to this discussion. 
We deliver a tool which gonna be used by clients. How they use it is up to them. When I take a look at the current available data
{:count=>11, :period=>60, :limit=>20} the time_to_reset is missing. 

I would even say that {:time_to_reset=>12 , :remaining_requests=>6} would be enough instead of those 3 above - all clients need to know how many calls they can trigger in a specific timeframe.
remaining_requests can be inferred from data above, but time_to_reset - not

ktheory:
@wpolicarpo, @twiduch:

For background, please keep in mind that Rack::Attack's primary use-case is preventing harm from *abusive clients*. Since exposing X-RateLimit headers doesn't clearly help stop abusive clients, it won't be the default.

But a secondary use-case is assisting well-behaved clients use a site appropriately. That's what these `X-RateLimit-*` headers are for.

Turns out we *can* determine `X-RateLimit-Reset` from the match_data's `period` and `Time.now`. E.g.:

```ruby
now = Time.now
match_data = env['rack.attack.match_data']

{ 'X-RateLimit-Reset' => (Time.now + (match_data[:period] - now.to_i % match_data[:period])).to_s }
```

In #193, I added a section to the README for how to customize you're throttle_response to set X-RateLimit-* headers. Please [check out this section](https://github.com/ktheory/rack-attack/blob/557f88daa1cd2b8fac2c406229e6e5d9713a19c5/README.md#x-ratelimit-headers-for-well-behaved-clients)

Given that this functionality is already possible using throttle_response, I'm going to close this out.

PR #193, opened 2016-08-03
README: add example for returning X-RateLimit-* headers

ktheory:
This addresses a use case from #191 of wanting to return rate limit information (especially `X-RateLimit-Reset`) for well-behaved clients.

Related: #29

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,18 +230,39 @@ Rack::Attack.blocklisted_response = lambda do |env|
 end
 
 Rack::Attack.throttled_response = lambda do |env|
-  # name and other data about the matched throttle
-  body = [
-    env['rack.attack.matched'],
-    env['rack.attack.match_type'],
-    env['rack.attack.match_data']
-  ].inspect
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -230,19 +230,40 @@ Rack::Attack.blocklisted_response = lambda do |env|
...
+Here's an example response that includes conventional `X-RateLimit-*` headers:
+
+```ruby
+Rack::Attack.throttled_response = lambda do |env|
+  now = Time.now
+  match_data = env['rack.attack.match_data']
+
+  headers = {
+    'X-RateLimit-Limit' => match_data[:limit].to_s,
+    'X-RateLimit-Remaining' => 0.to_s, # or [match_data[:limit] - match_data[:count], 0].max
```
tiegz: maybe add `to_s` after `max` too?

ktheory: @tiegz 👍 I can just simplify this as `'0'`.

PR #196, opened 2016-08-05
Custom named responses

madlep:
Hey there

Here's a PR to allow custom responses to be returned depending on the blocklist/throttle that matched. We needed this as we wanted to be able to return a slightly different response depending on what action Rack::Attack took, and jamming all the logic into one custom response was getting messy.

So this lets you do something like:

```ruby
Rack::Attack.set_named_blocklisted_response("geoblocked", lambda do |env|
  [451, {'Content-Type' => 'text/plain'}, ["lawyer says no\n"]]
end)

Rack::Attack.blocklist("geoblocked") do |req|
  ["Banstralia", "New Zeabanned"].include? MyCountryDB.find_by_ip(req.ip)
end
```

specifically:
- add `Rack::Attack.set_named_blocklisted_response` and `Rack::Attack.set_named_throttled_response` that takes a throttle/blocklist name and a response lambda and only reponds with it if that name match is triggered. If no named response is triggered, it uses the default response (`.blocklisted_response=` and `.throttled_response=` work the same as before and now set the default responses)
- turn `rack.attack...` rack response env names into constants in `Rack::Attack` to remove magic strings throughout code
- add `ResponseRegistry` class to handle response management. Delegate to that from `Rack::Attack`
- add tests for new logic
- add tests for `throttle` (which didn't seem to exist already)

Have a look and let me know your thoughts. Happy to rework if this doesn't fit.


madlep:
Bump. Any thoughts on this? Happy to rework things if it doesn't fit in it's current form.

ktheory:
Hi @madlep!

Thank you for taking the time to put this PR together and make rack-attack more useful.

I grant that managing different responses in a single lambda or method can get messy.

But rather than build a custom registry solution into rack-attack, keep in mind that theses responses are themselves rack apps.

A more elegant & generic solution would be to make a custom class with a 'call' method that encapsulates your logic for checking env['rack.attack.*'].

I'd welcome a PR documenting an example of such a response class.

madlep:
Thanks for that @ktheory I'll have a play with that approach

ktheory:
@madlep 👌 I'll close this for now then. Feel free to re-open with the rack app approach.

PR #206, opened 2016-10-28
Add LeakyBucket-based throttling

### LINKED ISSUE #205: Add leaky-bucket algorithm
The [Leaky Bucket Algorithm](https://en.wikipedia.org/wiki/Leaky_bucket) seems like a more natural algorithm to throttle requests. I've integrated this in a PHP port of rack-attack and authored it as a individual plugin: [jeroenvisser101/LeakyBucket](https://github.com/jeroenvisser101/LeakyBucket). I'd love to contribute this, as Ruby code, to rack-attack. Is this something that would fit in the vision of rack-attack?

jeroenvisser101:
Using the leaky bucket algorithm, you can get a better distrubution that allows a more human behaviour and that should (in theory) decrease false-positives.

It does that because it throttles differently. Let's take the old one, and concider you have a limit of 3 for 5 mins. If the user then did 3 requests, it'll have to wait for the next 5-min slot to do more requests.

Using the leaky bucket algorithm, doing 3 requests adds 3 drops in the bucket, and the bucket is full. The bucket then leaks consistently, and when leaked enough, it'll allow 1 more request (and then it will be full again).

[Shopify uses and explains this too here](https://help.shopify.com/api/guides/api-call-limit).

We might want to add a custom throttled_response for the leaky bucket algorithm, so that it has the limits in there too.

Fixes #205.

### DIFF
--- a/lib/rack/attack/throttle_with_leaky_bucket.rb
+++ b/lib/rack/attack/throttle_with_leaky_bucket.rb
@@ -1,0 +1,68 @@
+module Rack
+  class Attack
+    class ThrottleWithLeakyBucket
+      MANDATORY_OPTIONS = [:capacity, :leak]
+      attr_reader :name, :capacity, :leak, :block, :type
+
+      def initialize(name, options, block)
+        @name, @block = name, block
+        MANDATORY_OPTIONS.each do |opt|
+          raise ArgumentError.new("Must pass #{opt.inspect} option") unless options[opt]
+        end
+        @capacity = options[:capacity]
+        @leak     = options[:leak]
+        @type     = options.fetch(:type, :throttle_with_leaky_bucket)
+      end
+
+      def cache
+        Rack::Attack.cache
+      end
+
+      def [](req)
+        discriminator = block[req]
+        return false unless discriminator
+
+        # Normalize blocks to values
+        current_capacity = normalize_block(capacity, req)
+        current_leak     = normalize_block(leak, req)
+
+        # Read the bucket data and unserialize it. We only update the bucket data
+        # if we've changed the value. We don't write to update the leaked amount
+        # since that can be calculated and since the TTL will remove the item when
+        # it has drained.
+        key = "#{name}:#{discriminator}"
+        bucket = LeakyBucket.unserialize(cache.read(key), current_capacity, current_leak)
+        throttled = bucket.full?
+        bucket.add(1) unless bucket.full?
+        store_bucket(key, bucket) if bucket.updated?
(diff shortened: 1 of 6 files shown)


jeroenvisser101:
Sorry, this doesn't work yet, just tried it out in my app, and it's not blocking. Will investigate & fix.

jeroenvisser101:
It works now—turns out not rounding the `:expires_at` will cause it not to be persisted at all. It's now being rounded up, and works fine!

jeroenvisser101:
@ktheory I've also added documentation to the README, if you have time, could you check this out? Meanwhile, I'll push this into staging/prod to try out.

lib/rack/attack/throttle_with_leaky_bucket.rb
```diff
@@ -0,0 +1,68 @@
...
+
+        # Read the bucket data and unserialize it. We only update the bucket data
+        # if we've changed the value. We don't write to update the leaked amount
+        # since that can be calculated and since the TTL will remove the item when
+        # it has drained.
+        key = "#{name}:#{discriminator}"
+        bucket = LeakyBucket.unserialize(cache.read(key), current_capacity, current_leak)
+        throttled = bucket.full?
+        bucket.add(1) unless bucket.full?
+        store_bucket(key, bucket) if bucket.updated?
```
ktheory: @jeroenvisser101 Since the cached value is read in line 34, and written in line 37, this implementation is not atomic, and can lead to uncounted requests.

jeroenvisser101:
Hey @ktheory, you're right, I didn't think about being atomic here.

I'll have a look tonight to see if I can make it atomic in a deterministic fashion. Thanks for checking this out!

jeroenvisser101:
So I can do the following:

1. Deterministic and atomic increases of the bucket's value. (Using `INCR` just like throttle)
2. Non-atomic timestamp updates, but since these don't change so much based on the amount of requests, so atomic updates will not make that much of a difference?

This will mean that for each client, 2 keys will be stored. I'll also change the TTL logic to be the amount of seconds it takes to drain the full capacity (rather than the current value), since it wouldn't be possible to update that in an atomic way.

ktheory:
Gonna close this out as it's stalled. Feel free to re-open with updates.

jeroenvisser101:
@ktheory yeah, I think this idea is a little dead, I've not been able to come up with a solution that's fully atomic, but it's semi-atomic. If I have time I'll set up a new PR, thanks for all the feedback so far!

PR #221, opened 2017-02-16
Adds support for Redis::DistributedStore

mezis:
`Redis::Distributed` doesn't support pipelining the increment and the resetting of the TTL.
This replaces it with a Lua script (which is supported when sharding).


grzuy:
Hi @mezis,

Thank you for you pull request.

I don't think we'll add this to rack-attack in the short term.

Hopefully, once https://github.com/kickstarter/rack-attack/issues/380 gets implemented, you will be able to turn this into an adapter for you and others to use it without needing to push it upstream.

That's the idea going forward, to let users define new integrations easily, without friction, instead of continue adding extra store integrations here and becoming a bottleneck.

Sorry for the delay in the response.

PR #222, opened 2017-02-20
Throw exception if Redis gem is < 3.0.0.

rfwatson:
The RackAttack gem requires Redis gem with a version greater than 3.x.

Previously the gem silently failed to work if an old Redis gem was installed.

This change will throw an exception if this prerequisite is missing, hopefully making detection of this problem easier.


srt32:
I'm curious what folks think about raising an exception versus logging a non-breaking warning instead.

Logging an error could be more flexible than an exception if there are existing installs that use older versions of the redis gem and have managed to work around the issue. WDYT?

rfwatson:
Am I right in thinking that printing the message to stdout is the right way to do this?

srt32:
What about something like

```
warn 'RackAttack requires Redis gem >= 3.0.0.' if major_version < 3
```

as is used in https://github.com/kickstarter/rack-attack/blob/master/lib/rack/attack.rb#L29?

rfwatson:
Fixed in 69ab514

lib/rack/attack/store_proxy/redis_store_proxy.rb
```diff
@@ -4,6 +4,13 @@ module Rack
   class Attack
     module StoreProxy
       class RedisStoreProxy < SimpleDelegator
+        def initialize(*args)
```
grzuy: This is not being called because there's already an `#initialize` method a couple of lines below, which is overriding this one.

lib/rack/attack/store_proxy/redis_store_proxy.rb
```diff
@@ -4,6 +4,13 @@ module Rack
...
+          major_version = Redis::VERSION.split('.').first.to_i
+          warn 'RackAttack requires Redis gem >= 3.0.0.' if major_version < 3
```
grzuy: Can we please use `Gem::Version` built-in version comparison instead of manually splitting the string?

Something like `Gem::Version.new(Redis::VERSION) < Gem::Version.new("3")` e.g.?

grzuy:
Did the changes myself.

Thanks again!

PR #244, opened 2017-10-02
Use the latest Rubies on Travis CI

hisas:
Ruby 2.1.10, 2.2.8 and 2.3.5 has been released and these versions are available on Travis CI.


frewsxcv:
addressed in https://github.com/kickstarter/rack-attack/pull/261

Issue #225, opened 2017-04-01
Not working for rails 5.0.1 api only app


vamsipavanmahesh:
I have put the rack-attack in the gem file. 
In the application.rb did config.middleware.use Rack::Attack
```
class Rack::Attack
  
  throttle('req/ip', :limit => 2, :period => 2.minutes) do |req|
    req.ip # unless req.path.start_with?('/assets')
  end

end
```
The above is config/intializers/rack-attack.rb

Is there something else I need to do here? I did the same thing with rails 4.2 app, it works fine there. But here it does not seem to work

pachonk:
Well what's broken? Is there an error message that gors with it?

vamsipavanmahesh:
Well with the above config, I should get 429 status code if I make more than 2 requests for 2 minutes. But it's not stopping the client from making more than 2 rrequests

pachonk:
Have you tried restarting your rails server?

vamsipavanmahesh:
Yes I did .. but still it's the same, do you mind telling if it's working for you.

pachonk:
After playing, this worked for me:

config/initializers/rack-attack.rb:
```ruby
class Rack::Attack
  Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new 

  throttle('api/ip', limit: 2, period: 2.minutes) do |req|
    req.ip
  end
end
```

pachonk:
Looks like the docs _do_ mention this, actually:

> Optionally configure the cache store for throttling or fail2ban filtering

It is worded weird though.

pachonk:
@vamsipavanmahesh usually you want to state that the above resolved your issue before you close it, so people stuffing the internet at a later date can use the same technique.
(The only reason I know this resolved your issue is because I talked to you outside of GitHub)

vamsipavanmahesh:
Okies, lesson learnt .. the above solution worked!!

gambhirmakwana1992:
> After playing, this worked for me:
> 
> config/initializers/rack-attack.rb:
> 
> ```ruby
> class Rack::Attack
>   Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new 
> 
>   throttle('api/ip', limit: 2, period: 2.minutes) do |req|
>     req.ip
>   end
> end
> ```

I Implement the solution and working fine for me
Thanks

PR #247, opened 2017-10-31
Make it clear that the cache store can be required

frenkel:
In order to prevent issues such as #225

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ end
 *Tip:* The example in the wiki is a great way to get started:
 [Example Configuration](https://github.com/kickstarter/rack-attack/wiki/Example-Configuration)
 
-Optionally configure the cache store for throttling or fail2ban filtering:
+Configure the cache store if you want to use throttling or fail2ban filtering:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


lmansur:
We use `Rails.cache` as a fall back in case you didn't configure `Rack::Attack::Cache.store`. Depending on your Rails version, `Rails.cache` in development environment uses `NullStore`, which doesn't actually store anything.

I think we could add some kind of notice regarding this possible case to avoid unnecessary debugging.

grzuy:
Agree.

E.g. https://github.com/kickstarter/rack-attack/pull/274 was an attempt to aid users in debugging store config errors, but there's definitely room for more improvement. Either by giving better warning/errors and/or clearer docs.

grzuy:
Created a new issue to track that need: https://github.com/kickstarter/rack-attack/issues/300

grzuy:
Closing this PR for now given the conversation stalled, plus there were other changes which were trying to solve the same confusion as part of https://github.com/kickstarter/rack-attack/issues/300.

@frenkel Thank you for trying to contribute anyway. Let me know if you still think what you were trying to solve isn't yet covered, so we can continue discussing how to do that.

PR #248, opened 2017-11-06
add require lines in README.md file

amirshadaab:


README.md
```diff
@@ -30,13 +30,15 @@ For Rails 3+ apps:
 
 ```ruby
 # In config/application.rb
+require "rack/attack"
```
grzuy: I think normally you don't need this require line in rails apps, given that by convention there's an auto-generated line that calls `Bundler.require` for all gems listed in the `Gemfile`.

grzuy:
Hi @amirshadaab, thanks for your contribution.

Were you able to see my review comment above?

amirshadaab:
@grzuy yes. I've updated the PR. thanks!

PR #260, opened 2018-01-17
Update docs to use new ruby hash syntax

grzuy:
So that gem users don't get the old hash syntax when copy-pasting examples into their app code.


grzuy:
Thank you!

PR #264, opened 2018-01-23
Drop support for unmaintaned Rails 3

grzuy:


frewsxcv:
dropping rails 3 sounds reasonable to me! 👋

PR #273, opened 2018-01-31
Don't make README examples depend on activesupport

grzuy:
So that users pasting these examples don't get

```
undefined method `seconds' for 5:Integer
```

if they're not using activesupport in their apps


frewsxcv:
sounds reasonable to me, thanks!

grzuy:
😃 👍

PR #274, opened 2018-02-01
Help users understand more clearly when the store is misconfigured

grzuy:
Raises more semantic errors instead of blowing up with

```
undefined method `increment' for nil:NilClass
```


grzuy:
Helps with https://github.com/kickstarter/rack-attack/issues/300

PR #280, opened 2018-02-06
[Fixes #213] Add support for the redis gem

### LINKED ISSUE #213: rack-attack-4.3.1 RedisStoreProxy does not accept plain redis
We are experiencing breakage after upgrading from `rack-attack-4.2.0` to `rack-attack-4.3.1`, due to what appears to be an API incompatibility with `redis`:

```
app error: wrong number of arguments (4 for 3) (ArgumentError)
[...]/bundled_gems/ruby/2.1.0/gems/redis-3.2.0/lib/redis.rb:689:in `setex'
[...]/bundled_gems/ruby/2.1.0/gems/rack-attack-4.3.1/lib/rack/attack/store_proxy/redis_store_proxy.rb:22:in `write'
[...]/bundled_gems/ruby/2.1.0/gems/rack-attack-4.3.1/lib/rack/attack/cache.rb:53:in `do_count'
[...]/bundled_gems/ruby/2.1.0/gems/rack-attack-4.3.1/lib/rack/attack/cache.rb:19:in `count'
[...]/bundled_gems/ruby/2.1.0/gems/rack-attack-4.3.1/lib/rack/attack/throttle.rb:27:in `[]'
[...]/bundled_gems/ruby/2.1.0/gems/rack-attack-4.3.1/lib/rack/attack.rb:58:in `block in throttled?'
[...]/bundled_gems/ruby/2.1.0/gems/rack-attack-4.3.1/lib/rack/attack.rb:57:in `each'
[...]/bundled_gems/ruby/2.1.0/gems/rack-attack-4.3.1/lib/rack/attack.rb:57:in `any?'
[...]/bundled_gems/ruby/2.1.0/gems/rack-attack-4.3.1/lib/rack/attack.rb:57:in `throttled?'
[...]/bundled_gems/ruby/2.1.0/gems/rack-attack-4.3.1/lib/rack/attack.rb:102:in `call'

```

Previously, we were able to pass an instance of `Redis` from version `3.2.0` of the `redis` gem
to `Rack::Attack`, which satisfies the version constraint provided in the [README](https://github.com/kickstarter/rack-attack/blob/v4.3.1/README.md):

```rb
Rack::Attack.cache.store = Rack::Attack::StoreProxy::RedisStoreProxy.new(
  Redis.new(some_redis_config)
)
```

The above configuration appears to have been recommended by others in a previous [issue](https://github.com/kickstarter/rack-attack/issues/102#issuecomment-66176688).

However, after [this commit](https://github.com/kickstarter/rack-attack/commit/dc1b6bf687bbb6c93a3f5b790eb9a0083a1e6bc0#diff-3d0e57d4b76fc949e3de00ea0a5bed8f), it appears that the contract of `RedisStoreProxy` was changed to now depend on an instance of `Redis::Store` and not `Redis`, as only `Redis::Store` from the `redis-store-1.1.5` gem
supports the fourth `raw: true` parameter introduced by the aforementioned commit.

This issue has been reported previously (see https://github.com/kickstarter/rack-attack/issues/136 and https://github.com/kickstarter/rack-attack/issues/190). 

It appears that `rack-attack` now depends on `redis-store` extensions to the interface, which is breaking older configurations that pass a plain `redis` instance. Can `redis` be preserved as a valid store for the `RedisStoreProxy`, to support these pre-existing configurations?

Thanks!

bfad:
While a cache-store proxy exists for the redis-store gem, no such proxy existed for using the redis gem itself. This fills that gap by adding such a proxy.

Resolves kickstarter/rack-attack#190

### DIFF
--- a/lib/rack/attack/store_proxy.rb
+++ b/lib/rack/attack/store_proxy.rb
@@ -1,10 +1,10 @@
 module Rack
   class Attack
     module StoreProxy
-      PROXIES = [DalliProxy, MemCacheProxy, RedisStoreProxy].freeze
+      PROXIES = [DalliProxy, MemCacheProxy, RedisStoreProxy, RedisProxy].freeze
 
       ACTIVE_SUPPORT_WRAPPER_CLASSES = Set.new(['ActiveSupport::Cache::MemCacheStore', 'ActiveSupport::Cache::RedisStore']).freeze
-      ACTIVE_SUPPORT_CLIENTS = Set.new(['Redis::Store', 'Dalli::Client', 'MemCache']).freeze
+      ACTIVE_SUPPORT_CLIENTS = Set.new(['Redis::Store', 'Redis', 'Dalli::Client', 'MemCache']).freeze
 
       def self.build(store)
         client = unwrap_active_support_stores(store)
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -13,6 +13,7 @@ class Rack::Attack
   autoload :DalliProxy,      'rack/attack/store_proxy/dalli_proxy'
   autoload :MemCacheProxy,   'rack/attack/store_proxy/mem_cache_proxy'
   autoload :RedisStoreProxy, 'rack/attack/store_proxy/redis_store_proxy'
+  autoload :RedisProxy,      'rack/attack/store_proxy/redis_proxy'
   autoload :Fail2Ban,        'rack/attack/fail2ban'
   autoload :Allow2Ban,       'rack/attack/allow2ban'
   autoload :Request,         'rack/attack/request'
--- a/lib/rack/attack/store_proxy/redis_proxy.rb
+++ b/lib/rack/attack/store_proxy/redis_proxy.rb
@@ -1,0 +1,44 @@
+require 'delegate'
+
+module Rack
+  class Attack
+    module StoreProxy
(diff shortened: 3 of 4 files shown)


lib/rack/attack/store_proxy.rb
```diff
@@ -1,10 +1,10 @@
 module Rack
   class Attack
     module StoreProxy
-      PROXIES = [DalliProxy, MemCacheProxy, RedisStoreProxy].freeze
+      PROXIES = [DalliProxy, MemCacheProxy, RedisStoreProxy, RedisProxy].freeze
 
       ACTIVE_SUPPORT_WRAPPER_CLASSES = Set.new(['ActiveSupport::Cache::MemCacheStore', 'ActiveSupport::Cache::RedisStore']).freeze
-      ACTIVE_SUPPORT_CLIENTS = Set.new(['Redis::Store', 'Dalli::Client', 'MemCache']).freeze
+      ACTIVE_SUPPORT_CLIENTS = Set.new(['Redis::Store', 'Redis', 'Dalli::Client', 'MemCache']).freeze
```
grzuy: Not sure we need `Redis` here for the proxy to work.
Do you think we do?

grzuy:
Made the change myself and merged.

Thanks again! :tada:

bfad:
Thanks for getting this in 🍾

PR #282, opened 2018-02-15
[Fixes #283] Provide epoch_time that was used for the request in the data set

### LINKED ISSUE #283: Possible race condition between counting requests and reporting additional data
We are using `rack-attack` and we want to report additional information for well-behaved clients. 

In particular we want to inform the client of what the limit is, when it gets reset, and how many requests are remaining. 

Consider some middleware that is doing something like this:
(Note: This is very similar to the sample code provided in the `X-RateLimit headers for well-behaved clients` section of the README, it just provides the data on every request, not just throttled ones)

```
class FooMiddleware
  def initialize(app)
    @app = app
  end

  def call(env)
    status, headers, body = @app.call(env)
    now = Time.now
    match_data = env['rack.attack.match_data']
    if match_data
      headers.merge!({
        'X-RateLimit-Limit' => match_data[:limit].to_s,
        'X-RateLimit-Remaining' => match_data[:limit] - match_data[:count],
        'X-RateLimit-Reset' => (now + (match_data[:period] - now.to_i % match_data[:period])).to_s
      })
    end
    [status, headers, body]
  end
end
```

This does the trick, but there is a race condition here. 

`Rack::Attack::Throttle#[]` uses `Rack::Attack::Cache#count` to count the number of requests it has seen so far. The cache uses the following code to generate which key it is using
```
      def key_and_expiry(unprefixed_key, period)
        epoch_time = Time.now.to_i
        # Add 1 to expires_in to avoid timing error: http://git.io/i1PHXA
        expires_in = (period - (epoch_time % period) + 1).to_i
        ["#{prefix}:#{(epoch_time / period).to_i}:#{unprefixed_key}", expires_in]
      end
```

So here is the race condition. 
1) the request enters the cache method at T1 (at the very end of the second)
2) that increments the counter (as expected)
3) the middleware that reports the headers uses `Time.now` but it is now T1+1 second
4) The `X-RateLimit-Remaining` has a value of the previous time frame, but the `X-RateLimit-Reset` has rolled over

To better illustrate this here is a series of headers that could happen

```
REQUEST 1 - This request processes as I would expect

HTTP/1.1 201 Created
Status: 201 Created
X-RateLimit-Limit: 300
X-RateLimit-Reset: 1518397920
X-RateLimit-Remaining: 133

REQUEST #2
This request processes as I would expect. Notice that the Reset date is the same and the RateLimit-Remaining has decreased by one.

HTTP/1.1 201 Created
Status: 201 Created
X-RateLimit-Limit: 300
X-RateLimit-Reset: 1518397920
X-RateLimit-Remaining: 132

REQUEST #3
This is where things get weird. You’ll notice that the Reset date has changed which should result in the Remaining value being reset, however, the remaining value did not reset.

HTTP/1.1 200 OK
Status: 200 OK
X-RateLimit-Limit: 300
X-RateLimit-Reset: 1518397980
X-RateLimit-Remaining: 131

REQUEST #4
Now things are as they should be. The reset and remaining are now in sync.

HTTP/1.1 201 Created
Status: 201 Created
X-RateLimit-Limit: 300
X-RateLimit-Reset: 1518397980
X-RateLimit-Remaining: 299
```

A solution here would be to also expose the time that was used in the match data so that anything that has access to the data can operating in the same timeframe that rack attack believes. 

I've written up a possible PR that would provide this behaviour here:
https://github.com/kickstarter/rack-attack/pull/282

doliveirakn:
Aiming to solve https://github.com/kickstarter/rack-attack/issues/283

Add an reader for the epoch_time variable in the cache so that it can also be returned in the data from the throttle.

This is allows access to the same time that the cache uses for the count. This can be important for clients that want to provide rate limit information for well-behaved clients

### DIFF
--- a/lib/rack/attack/throttle.rb
+++ b/lib/rack/attack/throttle.rb
@@ -26,12 +26,15 @@ module Rack
         current_limit  = limit.respond_to?(:call) ? limit.call(req) : limit
         key            = "#{name}:#{discriminator}"
         count          = cache.count(key, current_period)
+        epoch_time     = cache.last_epoch_time
 
         data = {
           :count => count,
           :period => current_period,
-          :limit => current_limit
+          :limit => current_limit,
+          :epoch_time => epoch_time
         }
+
         (req.env['rack.attack.throttle_data'] ||= {})[name] = data
 
         (count > current_limit).tap do |throttled|
--- a/README.md
+++ b/README.md
@@ -254,13 +254,13 @@ Here's an example response that includes conventional `X-RateLimit-*` headers:
 
 ```ruby
 Rack::Attack.throttled_response = lambda do |env|
-  now = Time.now
   match_data = env['rack.attack.match_data']
+  now = match_data[:epoch_time]
 
   headers = {
     'X-RateLimit-Limit' => match_data[:limit].to_s,
     'X-RateLimit-Remaining' => '0',
-    'X-RateLimit-Reset' => (now + (match_data[:period] - now.to_i % match_data[:period])).to_s
+    'X-RateLimit-Reset' => (now + (match_data[:period] - now % match_data[:period])).to_s
   }
 
   [ 429, headers, ["Throttled\n"]]
@@ -271,7 +271,7 @@ end
 For responses that did not exceed a throttle limit, Rack::Attack annotates the env with match data:
 
 ```ruby
-request.env['rack.attack.throttle_data'][name] # => { :count => n, :period => p, :limit => l }
+request.env['rack.attack.throttle_data'][name] # => { :count => n, :period => p, :limit => l, :epoch_time => t }
 ```
 
 ## Logging & Instrumentation
--- a/lib/rack/attack/cache.rb
+++ b/lib/rack/attack/cache.rb
@@ -3,6 +3,7 @@ module Rack
     class Cache
 
       attr_accessor :prefix
+      attr_reader :last_epoch_time
 
       def initialize
         self.store = ::Rails.cache if defined?(::Rails.cache)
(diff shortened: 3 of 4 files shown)


lmansur:
Hi @doliveirakn, thank you for implementing this PR.

Could you please write a test case for the issue you reported? This ensures the Gem is actually fixing your exact problem, helps document this specific use case and catches any future regression.

doliveirakn:
@lmansur I thought about a test case for the specific issue that was reported and I don't see a good way to write a spec for that. 

The reason being is that the race condition comes from the fact that the time used in the cache is not exposed. 

In the issue (https://github.com/kickstarter/rack-attack/issues/283) I have code that looks like this:

```
class FooMiddleware
  def initialize(app)
    @app = app
  end

  def call(env)
    status, headers, body = @app.call(env)
    now = Time.now
    match_data = env['rack.attack.match_data']
    if match_data
      headers.merge!({
        'X-RateLimit-Limit' => match_data[:limit].to_s,
        'X-RateLimit-Remaining' => match_data[:limit] - match_data[:count],
        'X-RateLimit-Reset' => (now + (match_data[:period] - now.to_i % match_data[:period])).to_s
      })
    end
    [status, headers, body]
  end
end
```

This PR as it stands does not fix the problem. However it does let code be written like:
```
class FooMiddleware
  def initialize(app)
    @app = app
  end

  def call(env)
    status, headers, body = @app.call(env)
    match_data = env['rack.attack.match_data']
    now = match_data[:epoch_time]
    if match_data
      headers.merge!({
        'X-RateLimit-Limit' => match_data[:limit].to_s,
        'X-RateLimit-Remaining' => match_data[:limit] - match_data[:count],
        'X-RateLimit-Reset' => (now + (match_data[:period] - now % match_data[:period])).to_s
      })
    end
    [status, headers, body]
  end
end
```

The tests now assert that the `epoch_time` is now available in the match_data. So long as that contract is unheld (and it is now included in the `X-RateLimit headers for well-behaved clients` section of the Readme) users should be able to avoid this race condition in the future. I believe that to be sufficient. if you don't, would you be able to point to see what kind of test you would like to see and I will happily add it.

lmansur:
I think you are right, the gem shouldn't really care about the user's implementation.

PR looks fine for me. :+1:

lib/rack/attack/throttle.rb
```diff
@@ -26,12 +26,15 @@ def [](req)
         current_limit  = limit.respond_to?(:call) ? limit.call(req) : limit
         key            = "#{name}:#{discriminator}"
         count          = cache.count(key, current_period)
+        epoch_time     = cache.last_epoch_time
```
grzuy: I wonder whether making the epoch time an argument to the `count` method might end up feeling a bit cleaner... Instead of keeping the last epoch time in the cache.

Not a blocker for merging though, just a consideration for a possible future refactor :-)

Issue #219, opened 2017-02-10
Looking for maintainers


ktheory:
:wave: Welcome friends of Rack::Attack,

This project has grown to be more than I can support by myself. I'm looking to add some maintainers.

### What I'm looking for from you:

1. Be a helpful and welcoming voice when responding to GitHub issues and pull requests. Today's newbies are tomorrow's core contributors.
2. Exercise clear judgement about what should and should not be a feature of Rack::Attack, and strive to improve Rack::Attack's code quality.

### Roadmap

Feel like jumping in and writing code & documentation? Here are some suggestions for things to work on:

- Make it easier for new users to debug misconfigured cache stores & rules
- Reconsider the proxy pattern used to smooth over differences in cache stores. Is there a simpler, more robust way to support disparate stores?
- Keep Rack::Attack up-to-date with community conventions, and test against with the latest gem dependencies & ruby versions

Please leave a comment below if you're interested!

oyinloye-ai:
Hi ktheory, I am a rookie rail developer but will be interested in being a maintainer. Please buzz me on the way forward. Thanks.

ktheory:
@oyinloye-ai Awesome, and welcome. 👋 

The next step would be to chime in on the open issues with fixes, workarounds, or requests for clarifying information.

giubueno:
Hey pick me! I love this gem.

tjnet:
I love this gem too 😄 
I want to contribute this project!

jasonfb:
👍

buck3000:
Love the gem, definitely interested

PR #284, opened 2018-02-20
Fix several code style offenses

lmansur:
As @ktheory suggested in #219, this is an attempt keep the project up to date with the [community style guide](https://github.com/bbatsov/ruby-style-guide).

This is a good space to discuss what to adopt and what not to adopt from the guide. Should we set some guidelines or just use Rubocop? Or just ignore style all together?

I feel like following some kind of guideline helps the code be more consistent and less of a hassle to understand.

### DIFF
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -21,11 +21,10 @@ class Rack::Attack
   autoload :Request,         'rack/attack/request'
 
   class << self
-
     attr_accessor :notifier, :blocklisted_response, :throttled_response
 
     def safelist(name, &block)
-      self.safelists[name] = Safelist.new(name, block)
+      safelists[name] = Safelist.new(name, block)
     end
 
     def whitelist(name, &block)
@@ -34,7 +33,7 @@ class Rack::Attack
     end
 
     def blocklist(name, &block)
-      self.blocklists[name] = Blocklist.new(name, block)
+      blocklists[name] = Blocklist.new(name, block)
     end
 
     def blacklist(name, &block)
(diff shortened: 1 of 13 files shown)


grzuy:
Hi @lmansur 

> This is a good space to discuss what to adopt and what not to adopt from the guide. Should we set some guidelines or just use Rubocop? Or just ignore style all together?

If adopting a styleguide, then i think it probably makes sense to use something like rubocop so it's easy to continue to enforce it for future changes

> I feel like following some kind of guideline helps the code be more consistent and less of a hassle to understand

I agree

lmansur:
With that in mind, should we add a section on Rubocop to the `Contributing` section of the README?

What about adding a static code analyzer such as CodeClimate or Hound?

lmansur:
The PR is done.

@frewsxcv what are your thoughts on setting up a CodeClimate hook to ensure we are always merging quality code?

lmansur:
Fixed conflicts.

grzuy:
Did you run any `rubocop` command to spot these offenses? Was that run with any particular rubocop config file that you have?

Thanks.

lmansur:
@grzuy I just ran `rubocop` in the shell. It is using the default config, I didn't overwrite it.

grzuy:
@lmansur Gotcha, however if i run `rubocop` in this branch i still get a lot of offenses. Does the same happen to you?

grzuy:
What about pausing here for a moment, and discuss what we actually want to do, how and why in this issue i opened: https://github.com/kickstarter/rack-attack/issues/302? Then we can either move on with this or open new PR depending on what we want out of it.

:-)

lmansur:
>Gotcha, however if i run rubocop in this branch i still get a lot of offenses. Does the same happen to you?

@grzuy Yes, there are still a lot of offenses in the project. My intention here was not to fix all of them at once. I wanted to start with only a few so the reviewers could weigh in on them.

I will move the discussion to the issue #302.

lmansur:
@grzuy I'm closing this PR for now in favor of #330.

grzuy:
Sounds good, thanks! :-)

PR #285, opened 2018-03-06
Mention 'match_discriminator' in README.md

alextaylor000:
Thanks for your work on this gem! I have a small change to propose to the README. I was looking to return the cache key in a throttle response, and noticed that the `rack.attack.match_discriminator` environment key provides this. Thought it would be useful to mention as another example of what you can return in a throttle response.


grzuy:
Thanks Alex!

PR #287, opened 2018-03-10
Return time remaining to reset in Retry-After

### LINKED ISSUE #209: throttled requests seem to have a fixed retry after value independently of the time left to expire
When a request is throttled, the retry-after header value seems to be fixed:the throttle period, independently of the time left for the throttle to expire

lmansur:
fixes #209


grzuy:
Hi @lmansur, thank you for taking the time to work on this.

I was catching up today with some of the old issues in the issue tracker, and i am still not sure whether we want this or not, discussion in #209.

Issue #293, opened 2018-03-13
Improve acceptance tests coverage


grzuy:
Continue what started in https://github.com/kickstarter/rack-attack/pull/269, by writing more test cases that make assertions specifically about the behavior the user should expect the gem to provide to them, without any assertions about the internals or implementation details.

> [...] testing with respect to user needs, requirements [...] conducted to determine whether a system satisfies the acceptance criteria.
>
> from https://en.wikipedia.org/wiki/Acceptance_testing

---

User-facing features:

* [x] blocklist (#269)
* [x] safelist (#269)
* [x] throttle (#269)
* [x] allow2ban (#304)
* [x] fail2ban (#303)
* [x] tracks (#294)
* [x] custom throttled response (#298)
* [x] custom blocked response (#298)
* [x] cache store config for throttle (#297)
* [x] cache store config for allow2ban (#305)
* [x] cache store config for fail2ban (#305)
* [x] cache store config when Rails present (#307)
* [x] throttle's Retry-After HTTP response header (#306)
* [x] throttle's limit argument as a proc (#309)
* [x] throttle's period argument as a proc (#309)
* [x] ability to extend request object (#310)
* [x] blocklisted_response access to match data (#311)
* [x] throttled_response access to match_data (#311)
* [x] notifications (#319)

PR #294, opened 2018-03-13
Acceptance test Rack::Attack#track

grzuy:
part of #293

### DIFF
--- a/spec/acceptance/track_spec.rb
+++ b/spec/acceptance/track_spec.rb
@@ -1,0 +1,27 @@
+require_relative "../spec_helper"
+
+describe "#track" do
+  it "notifies when track block returns true" do
+    Rack::Attack.track("ip 1.2.3.4") do |request|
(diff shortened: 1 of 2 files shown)


lmansur:
I feel like both tests could be in the same file but within different `describe` blocks, since they are testing the behavior of the same method but with different options. What do you think?

grzuy:
Despite being exposed to the user with the same public method (`#track`), i think they have enough differences in purpose that it becomes helpful to think of them in isolation. E.g. the block argument has a different interpretation in each case. In one, is intended to flag when to track or not, in the other it returns the discriminator for the throttle being tracked.

I considered what you're proposing when i was writing these tests, but i preferred to err on the side of keeping them separate, so that both use cases can be thought of in isolation more easily.

Thank you for you feedback :-) :+1:

lmansur:
That makes sense :+1:

PR #312, opened 2018-03-23
support sub second throttling

amirshadaab:
```
2.2.3 :016 > 0.5.second.to_f
 => 0.5
2.2.3 :017 > 1.5.minute.to_f
 => 90.0
2.2.3 :018 >
```

This change will allow users to pass fractional values as the period.

Eg: 
```ruby
throttle("api_client_key", :limit => 2, :period => 0.5.second) do |request|
  #request.env["ip_address"]
end

```

I tested this change locally on one of my projects and it seems to be working fine.


amirshadaab:
@grzuy can you please help me out with this PR? If this is a feature you are not looking to add to this gem, I'll close this PR.

amirshadaab:
@grzuy I really wanted to use this feature in my project. I know a lot of people at my company who would love this feature. That is the reason why I was trying to implement it.

grzuy:
I see.

So the goal for you is trying to use

```ruby
throttle("api_client_key", limit: 2, period: 0.5)
```

instead of doing

```ruby
throttle("api_client_key", limit: 4, period: 1)
```
?

If so, do you think there's a significant difference in the amount of protection you're providing to your API with either implementation?

amirshadaab:
@grzuy That is correct. Just to be clear, the 0.5 second vs the 1 second granularity matters. I've had experiences where someone tries to call our service 1000 times a second. They do so by concentrating their requests on the top of a second (Please see image). So almost all 1000 requests come in in a few milliseconds which ends up degrading the performance of the service. I am trying to avoid that. I want to make sure the users spread out their requests over the second or they get an error code. That way we can prevent our service's request time from degrading.

image

grzuy:
How are 1000 requests a problem if you have a "4 per second" throttle configured?
Wouldn't the 5th and subsequent requests be blocked?

amirshadaab:
Here are the conditions I want satisfied
**1st condition:** I want to allow only 1000 requests per second
**2nd condition:** I want to make sure that those 1000 allowed requests are spread out over the second and are not concentrated on top of a second

With the current implementation where 1 second is the lowest I can go, I will not be able to ensure that the second condition is met.

I can only do it if I could reduce the granularity to less than a second. Then I can specify that only 200 requests are allowed per 0.2 second. This will obviously satisfy the first condition (as it did before) and also make sure the requests are more spread out.

I hope this makes sense.

grzuy:
That makes sense. Thank you for taking to time to explain! :-)

Moving forward, can you please try to fix existing test case is failing in this pull request, plus consider adding an additional test case for this particular feature in `spec/acceptance/throttling_spec.rb`?

Please let me know if you need any help or guidance.
Thanks!

grzuy:
Closing after 18 months of inactivity.
Let me know if you have intentions of finishing, so we can re-open.

Thank you very much anyway for your efforts.

PR #339, opened 2018-05-14
Replace `const_defined?` with `defined?`

brian-kephart:
[The bug leading to this code](https://bugs.ruby-lang.org/issues/14407) appears to be fixed.


grzuy:
Hi Brian,

Thanks a lot! :tada:

PR #340, opened 2018-05-14
Support ActiveSupport::RedisCacheStore

### LINKED ISSUE #281: Support RedisCacheStore (new in Rails 5.2)
After upgrading my app to Rails 5.2.0rc1, I'm occasionally getting the following error:
`Redis::CommandError: ERR value is not an integer or out of range`

Google says this error is caused by not passing the `raw` option when writing to Redis, then later incrementing the key. This appears to be fixed in this library for other forms of Redis caching, but not for the new ActiveSupport::Cache::RedisCacheStore.

Previous (working) config:
Ruby 2.5.0
Rails 5.1.4
ReadThisStore
Hiredis (driver)

Current (non-working) config:
Ruby 2.5.0
Rails 5.2.0rc1
RedisCacheStore
Hiredis (driver)

brian-kephart:
Fixes #281. Thanks for the good test suite, it prevented many problems. This PR passes Travis & Rubocop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+- Adds support for ActiveSupport::RedisCacheStore
+
 ## [5.2.0] - 2018-03-29
 
 ### Added
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -8,18 +8,19 @@ class Rack::Attack
   class MisconfiguredStoreError < StandardError; end
   class MissingStoreError < StandardError; end
 
-  autoload :Cache,           'rack/attack/cache'
-  autoload :Check,           'rack/attack/check'
-  autoload :Throttle,        'rack/attack/throttle'
-  autoload :Safelist,        'rack/attack/safelist'
-  autoload :Blocklist,       'rack/attack/blocklist'
-  autoload :Track,           'rack/attack/track'
-  autoload :StoreProxy,      'rack/attack/store_proxy'
-  autoload :DalliProxy,      'rack/attack/store_proxy/dalli_proxy'
-  autoload :MemCacheProxy,   'rack/attack/store_proxy/mem_cache_proxy'
-  autoload :RedisStoreProxy, 'rack/attack/store_proxy/redis_store_proxy'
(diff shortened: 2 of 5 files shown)


grzuy:
> Fixes #281.

Thanks!

> Thanks for the good test suite, it prevented many problems.

Thank a lot for the feedback. I especially appreciate you noticing that, given that making the test suite robust was the main focus a couple of months ago (https://github.com/kickstarter/rack-attack/issues/293). Any other feedback about ways to improve it more very welcomed!

> This PR passes Travis & Rubocop.

Thanks for checking rubocop also.

grzuy:
I want to code review and test this a bit more before merging. Will do as soon as I can. Will let you know.
Thanks again.

brian-kephart:
Sounds good. Let me know if you have questions, or if anything's unclear.

brian-kephart:
I change my initializer to:
```ruby
throttle('req/ip', :limit => 2, :period => 10) do |req|
  req.ip # unless req.path.start_with?('/assets')
end
```

...and now I can reproduce the error reliably. If I have the app running, then stop and restart Redis, the second request produces the error. Subsequent requests reproduce it until the 10sec timeout period expires.

Why? No idea.

brian-kephart:
I put the repro on github [here](https://github.com/brian-kephart/rack-attack-test).

brian-kephart:
Closing. See notes on #281.

morgoth:
@brian-kephart I'm just debugging issues with Rails 5.2 using built in Redis cache handler and rack attack keys in Redis that don't expire. It looks like your PR could fix it. Why it was closed?

brian-kephart:
The only errors I could reliably reproduce were solved another way, see #281 as mentioned above. I haven't seen these errors again despite weeks of use.

I think @grzuy would be open to merging this, but only with a reliable repro and a test. Current tests pass using the RedisCacheStore, so if it's really not working then there's something not covered by the tests.

grzuy:
@brian-kephart @morgoth I was able to add a test reproducing the issue, see commits in https://github.com/kickstarter/rack-attack/pull/350.

This PR can no longer be merged because the original repo was deleted. I am cherry-picking the commit anyways.

Thank you both!

brian-kephart:
Thanks @grzuy for your patience and persistence on this!

Issue #342, opened 2018-05-17
Acceptance test all supported cache store backends


grzuy:
(follow-up of https://github.com/kickstarter/rack-attack/issues/293)

__As__ rack-attack developers
__We want__ rack-attack integration with different cache store backends to be tested by acceptance test cases, which test from the gem user perspective
__So that__ we can be fairly confident that there are no user-facing regressions for each merged PR and release
__And__ is clearer which are the actually working cache store backends

---

Stores that are being tested in `spec/integration/rack_attack_cache_spec.rb`

* [x] `ActiveSupport::Cache::MemCacheStore` (via `activesupport`) (#343)
* [x] `ActiveSupport::Cache::RedisCacheStore` (via `activesupport`) (#341)
* [ ] `Dalli::Client` (via `dalli`) (#360)
* [ ] `ConnectionPool.new { Dalli::Client.new }` (via `connection_pool` and `dalli`) (#360)
* [ ] `ActiveSupport::Cache::DalliStore` (via `dalli`) (#360)
* [ ] `Redis::Store` (via `redis-store`) (#360)
* [ ] `ActiveSupport::Cache::RedisStore` (via `redis-activesupport`) (#360)
* [ ] `ActiveSupport::Cache::MemoryStore`  (via `activesupport`) (#360)

PR #343, opened 2018-05-18
Acceptance test use of ActiveSupport::Cache::MemCacheStore

grzuy:
part of #342


grzuy:
Thank you and the rest of the contributors for building this awesome gem in the first place :-)

PR #347, opened 2018-06-17
Add SemVer compatibility badge

greysteil:
First up, thanks for Rack::Attack. I use it in almost all my projects and love it.

This PR adds a SemVer compatibility badge that displays the percentage of CI runs that pass when updating Rack::Attack between SemVer compatible versions. Data comes from Dependabot (disclosure: I built it). The idea is to make it clear to people to know they can specify `~> 5.0` in their Gemfile, based on how well you guys have always observed SemVer.

Here's what the badge looks like:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=rack-attack&package-manager=bundler&version-scheme=semver)


lmansur:
That's awesome!

grzuy:
Hi @greysteil,

Thank your for the pull request.

I find intuitiveness, simplicity and obviousness to be important qualities for badges on top of the `README`. I value the fact that `README` readers can very quickly understand what the badge is measuring/telling just by skimming.

Not sure "semver stability" has these qualities, as of today, IMHO, despite trying to measure something which I find interesting.

Issue #190, opened 2016-07-14
redis gem only expects one argument in the `get` method


monfresh:
Hi. I've noticed other similar issues, but I couldn't get a definitive answer. When I implement a `Fail2Ban` filter, I get this error:

```
ArgumentError - wrong number of arguments (given 2, expected 1):
13:05:52 web.1    |   redis (3.3.0) lib/redis.rb:860:in `get'
13:05:52 web.1    |   /Users/moncefbelyamani/.rvm/rubies/ruby-2.3.1/lib/ruby/2.3.0/delegate.rb:83:in `method_missing'
13:05:52 web.1    |   rack-attack (4.4.1) lib/rack/attack/store_proxy/redis_store_proxy.rb:16:in `read'
13:05:52 web.1    |   rack-attack (4.4.1) lib/rack/attack/cache.rb:23:in `read'
13:05:52 web.1    |   rack-attack (4.4.1) lib/rack/attack/fail2ban.rb:26:in `banned?'
13:05:52 web.1    |   rack-attack (4.4.1) lib/rack/attack/fail2ban.rb:10:in `filter'
```

The README implies that I should be able to use the `redis` gem (v3.0.0+), but it looks like it's not supported because `RedisStoreProxy#read` passes in 2 arguments. From reading the other issues, it seems like I would need to add another dependency to my app for this work, namely the `redis-store` gem.

What is the purpose of the `raw: true` option, and is it essential? Could it be made optional? Is this a bug in the `redis` gem, or is it not required of it to accept a second argument? If the second argument is optional, then could rack-attack please update the `read` method such that it's compatible with the `redis` gem?

Thanks!

sgtpepper43:
I'm also seeing this issue with the `setex` command, you're passing in four arguments when it expects three. I don't know if it's _actually_ a problem, but I'm using bugsnag and it logs that error out quite a bit. Would appreciate a little clarification.

kgleong:
I'm also seeing a similar error with the `setex` method.

My rack-attack Redis store configuration in `application.rb`:

```
      redis = Redis.new(
        host: redis_config[:host],
        port: redis_config[:port],
        db: redis_config[:db],
      )

      Rack::Attack.cache.store =
        Rack::Attack::StoreProxy::RedisStoreProxy.new(redis)
```

ryantm:
[...]

Issue #136, opened 2015-07-12
4.3.0 fails to load config that works in 4.2.0


trinode:
**Update: forcing a downgrade to 4.2.0 from 4.3.0 solved the issue, no other changes. so this appears to be the latest version that doesn't work**

The file below is no longer working, I'm getting the error 

ArgumentError (wrong number of arguments (2 for 1)):
  config/initializers/rack_attack.rb:22:in `block in <class:Attack>'

(The line is the first Rack::Attack::Allow2Ban call)

```Ruby
class Rack::Attack
	Rack::Attack.cache.store = Rack::Attack::StoreProxy::RedisStoreProxy.new(Redis.current)

	Rack::Attack.whitelist('allow from localhost') do |req|
	  # Requests are allowed if the return value is truthy
	  '127.0.0.1' == req.ip
	end

	Rack::Attack.throttle('req/ip', :limit => 75, :period => 15.seconds) do |req|
	  	# If the return value is truthy, the cache key for the return value
	  	# is incremented and compared with the limit. In this case:
	  	#   "rack::attack:#{Time.now.to_i/1.second}:req/ip:#{req.ip}"
	  	#
	  	# If falsy, the cache key is neither incremented nor checked.
  		req.ip
	end

Rack::Attack.blacklist('user register flood') do |req|
   # `filter` returns false value if request is to your login page (but still
   # increments the count) so request below the limit are not blocked until
   # they hit the limit.  At that point, filter will return true and block.
   Rack::Attack::Allow2Ban.filter(req.ip, :maxretry => 20, :findtime => 1.minute, :bantime => 1.hour) do
     # The count for the IP is incremented if the return value is truthy.
     req.path == '/users' and req.post?
   end
 end

   Rack::Attack.blacklist('user login flood') do |req|
   # `filter` returns false value if request is to your login page (but still
   # increments the count) so request below the limit are not blocked until
   # they hit the limit.  At that point, filter will return true and block.
   Rack::Attack::Allow2Ban.filter(req.ip, :maxretry => 20, :findtime => 1.minute, :bantime => 1.hour) do
     # The count for the IP is incremented if the return value is truthy.
     req.path == '/users/sign_in' and req.post?
   end
	 end
  Rack::Attack.blacklist('admin login flood') do |req|
   # `filter` returns false value if request is to your login page (but still
   # increments the count) so request below the limit are not blocked until
   # they hit the limit.  At that point, filter will return true and block.
[...]

PR #348, opened 2018-06-18
Introduce a Redis proxy compatible with ~>redis-3.0.

### LINKED ISSUE #213: rack-attack-4.3.1 RedisStoreProxy does not accept plain redis

ryandv:
Resolves #213, #136, #190.

Root cause analysis is contained in #213.

In this new `RedisProxy` we just throw away the `redis-store` extensions to the plain `redis` interface. We have been using a similar approach in production for almost a year without issues. That being said, we only use the rate-throttling (and not the fail2ban) behaviour of rack-attack.

Let me know if you require further information or would like to make any amendments to the pull request.

### DIFF
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -18,6 +18,7 @@ class Rack::Attack
   autoload :DalliProxy,      'rack/attack/store_proxy/dalli_proxy'
   autoload :MemCacheProxy,   'rack/attack/store_proxy/mem_cache_proxy'
   autoload :RedisStoreProxy, 'rack/attack/store_proxy/redis_store_proxy'
+  autoload :RedisProxy,      'rack/attack/store_proxy/redis_proxy'
   autoload :Fail2Ban,        'rack/attack/fail2ban'
   autoload :Allow2Ban,       'rack/attack/allow2ban'
 
--- a/lib/rack/attack/store_proxy.rb
+++ b/lib/rack/attack/store_proxy.rb
@@ -1,7 +1,7 @@
 module Rack
   class Attack
     module StoreProxy
-      PROXIES = [DalliProxy, MemCacheProxy, RedisStoreProxy].freeze
+      PROXIES = [DalliProxy, MemCacheProxy, RedisStoreProxy, RedisProxy].freeze
 
       ACTIVE_SUPPORT_WRAPPER_CLASSES = Set.new(['ActiveSupport::Cache::MemCacheStore', 'ActiveSupport::Cache::RedisStore']).freeze
       ACTIVE_SUPPORT_CLIENTS = Set.new(['Redis::Store', 'Dalli::Client', 'MemCache']).freeze
--- a/lib/rack/attack/store_proxy/redis_proxy.rb
+++ b/lib/rack/attack/store_proxy/redis_proxy.rb
@@ -1,0 +1,29 @@
+require 'delegate'
+
+module Rack
+  class Attack
+    module StoreProxy
+      class RedisProxy < RedisStoreProxy
+        def self.handle?(store)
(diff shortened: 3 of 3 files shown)


morgoth:
I'm just debugging issues with non expiring keys in Rails 5.2 and built in Redis cache wrapper, and with this PR I found a working workaround.
Hopefully see it merged :+1:

morgoth:
BTW, looks like a duplicate of https://github.com/kickstarter/rack-attack/pull/280

grzuy:
Thank you @ryandv for taking the time to work on this pull request.

I am closing giving this is already fixed by https://github.com/kickstarter/rack-attack/pull/280.

PR #362, opened 2018-06-22
Test cache store dependencies in isolation

grzuy:


grzuy:
FTR, the reason why TravisCI is red is because by isolating cache store dependencies in each test suite run, we are successfully exposing/reproducing this existing bug https://github.com/kickstarter/rack-attack/issues/355.

PR #365, opened 2018-06-25
Remove support for unmaintained ruby 2.2

grzuy:


grzuy:
This is part of `v6.0.0`.

PR #374, opened 2018-07-06
Fix usage with connection-pooled MemCacheStore

jdelStrother:
We recently added `pool_size` to our Rails cache options:
```
config.cache_store = :mem_cache_store, *memcache_hosts, pool_size: 5
```

and discovered that Rack::Attack was no longer throttling.  (StoreProxy.build wasn't detecting the store as proxyable, so would return the bare ActiveSupport::Cache::MemCacheStore instance rather than a DalliProxy)

This PR refactors the ConnectionPool handling into a PoolProxy superclass, and makes StoreProxy.build aware of ConnectionPools.

WDYT?

### DIFF
--- a/lib/rack/attack/store_proxy.rb
+++ b/lib/rack/attack/store_proxy.rb
@@ -3,10 +3,7 @@
 module Rack
   class Attack
     module StoreProxy
-      PROXIES = [DalliProxy, MemCacheProxy, RedisStoreProxy, RedisProxy, RedisCacheStoreProxy].freeze
-
-      ACTIVE_SUPPORT_WRAPPER_CLASSES = Set.new(['ActiveSupport::Cache::MemCacheStore', 'ActiveSupport::Cache::RedisStore', 'ActiveSupport::Cache::RedisCacheStore']).freeze
-      ACTIVE_SUPPORT_CLIENTS = Set.new(['Redis::Store', 'Dalli::Client', 'MemCache']).freeze
+      PROXIES = [DalliProxy, MemCacheStoreProxy, MemCacheProxy, RedisStoreProxy, RedisProxy, RedisCacheStoreProxy].freeze
 
       def self.build(store)
         client = unwrap_active_support_stores(store)
@@ -17,15 +14,8 @@ module Rack
       def self.unwrap_active_support_stores(store)
         # ActiveSupport::Cache::RedisStore doesn't expose any way to set an expiry,
         # so use the raw Redis::Store instead.
-        # We also want to use the underlying Dalli client instead of ::ActiveSupport::Cache::MemCacheStore,
-        # and the MemCache client if using Rails 3.x
-
-        if store.instance_variable_defined?(:@data)
-          client = store.instance_variable_get(:@data)
-        end
-
-        if ACTIVE_SUPPORT_WRAPPER_CLASSES.include?(store.class.to_s) && ACTIVE_SUPPORT_CLIENTS.include?(client.class.to_s)
-          client
+        if store.class.name == 'ActiveSupport::Cache::RedisStore'
+          store.instance_variable_get(:@data)
         else
           store
         end
--- a/spec/acceptance/stores/active_support_mem_cache_store_pooled_spec.rb
+++ b/spec/acceptance/stores/active_support_mem_cache_store_pooled_spec.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require_relative "../../spec_helper"
+
+if defined?(::ConnectionPool) && defined?(::Dalli)
+  require_relative "../../support/cache_store_helper"
+  require "timecop"
+
+  describe "ActiveSupport::Cache::MemCacheStore (pooled) as a cache backend" do
+    before do
+      Rack::Attack.cache.store = ActiveSupport::Cache::MemCacheStore.new(pool_size: 2)
+    end
+
+    after do
+      Rack::Attack.cache.store.clear
+    end
+
+    it_works_for_cache_backed_features(fetch_from_store: ->(key) {
+      Rack::Attack.cache.store.read(key)
+    })
+  end
+end
--- a/spec/acceptance/stores/active_support_mem_cache_store_spec.rb
+++ b/spec/acceptance/stores/active_support_mem_cache_store_spec.rb
@@ -12,9 +12,9 @@ if defined?(::Dalli)
     end
 
     after do
-      Rack::Attack.cache.store.flush_all
+      Rack::Attack.cache.store.clear
     end
 
-    it_works_for_cache_backed_features(fetch_from_store: ->(key) { Rack::Attack.cache.store.get(key) })
+    it_works_for_cache_backed_features(fetch_from_store: ->(key) { Rack::Attack.cache.store.read(key) })
   end
 end
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -19,6 +19,7 @@ class Rack::Attack
   autoload :StoreProxy,           'rack/attack/store_proxy'
   autoload :DalliProxy,           'rack/attack/store_proxy/dalli_proxy'
   autoload :MemCacheProxy,        'rack/attack/store_proxy/mem_cache_proxy'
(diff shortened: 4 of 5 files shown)


spec/acceptance/store_proxy_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        assert_instance_of Rack::Attack::StoreProxy::RedisCacheStoreProxy, Rack::Attack::StoreProxy.build(store)
+      end
+    end
+    if defined?(::Redis) && defined?(::ActiveSupport::Cache::RedisCacheStore) && defined?(::ConnectionPool)
+      it "uses RedisCacheStoreProxy to handle plain ActiveSupport RedisCacheStore with a connection pool" do
+        store = ActiveSupport::Cache::RedisCacheStore.new(pool_size: 2)
+        assert_instance_of Rack::Attack::StoreProxy::RedisCacheStoreProxy, Rack::Attack::StoreProxy.build(store)
+      end
+    end
+  end
```
grzuy: Not sure I would place a test like this in the `spec/acceptance` folder. The idea behind test cases in that folder is to test from a gem user perspective, as much as possible, as opposed to internal implementation details of the gem.

It would be awesome if we could convert this into something that actually tests what you're actually reporting as the problem in the description, which is that "throttle isn't working when the store is a pooled `MemCacheStore`".

It seems what you need is something very very similar in essence to [spec/acceptance/stores/active_support_redis_cache_store_pooled_spec.rb](https://github.com/kickstarter/rack-attack/blob/master/spec/acceptance/stores/active_support_redis_cache_store_pooled_spec.rb) but for `MemCacheStore` instead of `RedisCacheStore`.

jdelStrother:
OK, working on it...

jdelStrother:
I've added the store acceptance spec, and moved the store_proxy_spec to spec/integration - not sure if that's the best location for it?  It could just go in the spec directory, but none of the specs in there rely on Dalli/Redis/etc.

spec/acceptance/stores/active_support_mem_cache_store_pooled_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    after do
+      Rack::Attack.cache.store.with { |client| client.flush_all }
+    end
+
+    it_works_for_cache_backed_features(fetch_from_store: ->(key) {
+      Rack::Attack.cache.store.with { |client| client.fetch(key) }
+    })
+  end
+end
```
grzuy: Excellent.

This acceptance test is perfect, thanks!

grzuy: I tested running this test case in `master`, and indeed it's failing.

spec/acceptance/stores/active_support_mem_cache_store_spec.rb
```diff
@@ -15,28 +15,6 @@
...
-        get "/", {}, "REMOTE_ADDR" => "1.2.3.4"
-      end
-
-      assert Rack::Attack.cache.store.get(key)
-
-      sleep 2.1
-
-      assert_nil Rack::Attack.cache.store.get(key)
-    end
+    it_works_for_cache_backed_features(fetch_from_store: ->(key) { Rack::Attack.cache.store.get(key) })
```
grzuy: Nice refactor :clap:

Thank you

spec/integration/store_proxy_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+    end
+    if defined?(::Redis) && defined?(::ActiveSupport::Cache::RedisCacheStore) && defined?(::ConnectionPool)
+      it "uses RedisCacheStoreProxy to handle plain ActiveSupport RedisCacheStore with a connection pool" do
+        store = ActiveSupport::Cache::RedisCacheStore.new(pool_size: 2)
+        assert_instance_of Rack::Attack::StoreProxy::RedisCacheStoreProxy, Rack::Attack::StoreProxy.build(store)
+      end
+    end
+  end
+end
```
grzuy: Feel free to leave this one out.

With the acceptance test case you added I think we are well covered. While this one is too coupled to the specifics of the implementation.

lib/rack/attack/store_proxy.rb
```diff
@@ -22,9 +22,10 @@ def self.unwrap_active_support_stores(store)
 
         if store.instance_variable_defined?(:@data)
           client = store.instance_variable_get(:@data)
+          client_class = PoolProxy.unwrap_connection_pool_class(client)
```
grzuy: Bummer, I was expecting we would need less to actually support `ActiveSupport::Cache::MemCacheStore.new(pool_size: 2)` as a store.

Do we really need to unwrap the `MemCacheStore` to make the acceptance test green?
Do you know what is causing `MemCacheStore#increment` not work by default?

grzuy:
Hi @jdelStrother,

Is there anything I can do to help you out with this PR?

jdelStrother:
Let me re-familiarize myself with the problem...

lib/rack/attack/store_proxy.rb
jdelStrother: If we fail to find a suitable Proxy class for the store, then Rack::Attack.cache.store is the unproxied `ActiveSupport::Cache::MemCacheStore` instance.

When we then call `store.increment` and `store.write` here: https://github.com/kickstarter/rack-attack/blob/master/lib/rack/attack/cache.rb#L57-L63

Normally our dalli proxy class would pass `raw: true` to the underlying instance, but because we don't have a proxy, a marshalled instance of `ActiveSupport::Cache::Entry` gets written to memcached.

The next call `increment` causes a dalli error of "Incr/decr on a non-numeric value".  This is caught by ActiveSupport's rescue handler (https://github.com/rails/rails/blob/4f09cff52c0ef771baac5f25c7c87a65068c0729/activesupport/lib/active_support/cache/mem_cache_store.rb#L202), and so returns nil

jdelStrother:
^ Sorry, had to change the PR's base branch before github would notice that I'd deleted that integration spec ¯\_(ツ)_/¯

lib/rack/attack/store_proxy.rb
grzuy: Gotcha.

Thank you very much for the details!

So, if what is missing is only `raw: true`, maybe we can turn the acceptance test from red to green just by implementing a very small `MemCacheStoreProxy` with a `#write` method wrapper, analogous to the existing [`RedisCacheStoreProxy#write`](https://github.com/kickstarter/rack-attack/blob/03b8ce6f9e44dad2385aa04e5e6d67231a087296/lib/rack/attack/store_proxy/redis_cache_store_proxy.rb#L31-L33)?

As much as we can avoid messing with the internals of the original cache store the better.

jdelStrother:
See what you think to https://github.com/kickstarter/rack-attack/compare/master...jdelStrother:memcache-store-proxy ?

With that change, we no longer try to unwrap ActiveSupport::Cache::MemcacheStore to get at the underlying dalli instance.  Instead, it's wrapped in a new MemcacheStoreProxy class, so the connection-pooling details are abstracted away by ActiveSupport.

AFAICT, `unwrap_active_support_stores` has always been useless for ActiveSupport::Cache::RedisCacheStore, since it doesn't have an `@data` instance variable.  So the only remaining ActiveSupport store to unwrap is RedisStore (https://github.com/redis-store/redis-activesupport/blob/master/lib/active_support/cache/redis_store.rb).

jdelStrother:
> Only question is if you sure we need to wrap #increment and #read also?

I'll check.  Was just getting the tests passing in that branch, I've not actually tried it in an app yet.

jdelStrother:
OK, I've force-updated this branch with the new MemcacheStoreProxy version. I've tested this in a Rails 5.2 app with a connection-pooled mem_cache_store cache, which seems to work fine.

And you're right, we don't need to override `increment` & `read` in the proxy.

grzuy:
FTR `#read` unnecessary for `RedisCacheStoreProxy` also. Removed in https://github.com/kickstarter/rack-attack/pull/382.

grzuy:
This is now released as part of `v5.4.1`.
Thanks again!

PR #376, opened 2018-07-31
Fix copy on safelist description

carlos-contreras:


grzuy:
Good catch!

Thanks a lot!

PR #377, opened 2018-08-16
Prefer to not use X- style headers for clarity

krainboltgreene:
The IETF decided the X- style custom headers have caused more headache than the value: https://tools.ietf.org/html/rfc6648 so this tool (which is very popular and much loved) shouldn't suggest using an old style.


krainboltgreene:
@grzuy Great catch, I've changed it.

PR #402, opened 2019-01-26
Add example: Match Actions in Rails

collimarco:
Instead of matching the URL with complex regex, it can be much easier to mach specific controller actions.

### DIFF
--- a/docs/advanced_configuration.md
+++ b/docs/advanced_configuration.md
@@ -91,3 +91,19 @@ Rack::Attack.blocklist('basic auth crackers') do |req|
   end
 end
 ```
+
+### Match Actions in Rails
+
+Instead of matching the URL with complex regex, it can be much easier to mach specific controller actions:
+
+```ruby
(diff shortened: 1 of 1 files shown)


docs/advanced_configuration.md
```diff
@@ -90,4 +90,20 @@ Rack::Attack.blocklist('basic auth crackers') do |req|
...
+
+Instead of matching the URL with complex regex, it can be much easier to mach specific controller actions:
+
+```ruby
+Rack::Attack.safelist('unlimited requests') do |request|
+  safelist = [
+    'controller#action',
+    'another_controller#another_action'
+  ]
+  route = (Rails.application.routes.recognize_path request.url rescue {}) || {}
```
grzuy: Which is the potential exception being rescued here?

collimarco: It can raise `ActionController::RoutingError` and maybe other exceptions. And yes this code is working fine in production.

grzuy:
Thank you!

PR #403, opened 2019-01-29
build: support bundler 2.0 for development

grzuy:


grzuy:
Also fixes broken build in `master` due to bundler version mismatch in CI.

PR #410, opened 2019-03-02
Code style updates

grzuy:
Continue what started in https://github.com/kickstarter/rack-attack/pull/330 and https://github.com/kickstarter/rack-attack/pull/338

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -49,8 +49,3 @@ Style/HashSyntax:
 
 Style/RedundantFreeze:
   Enabled: true
-
-# TODO
-# Remove cop disabling and fix offenses
-Lint/HandleExceptions:
-  Enabled: false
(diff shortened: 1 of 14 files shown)


grzuy:
@lmansur Would be great to have your feedback on this if you have any time.
Thank you! :-)

lmansur:
Thanks for the ping, @grzuy.

The PR looks good to me :+1:. I think we are in the right path of establishing a Rubocop file and committing to it.

Do you think it's worth setting up some kind of static code analysis (CodeClimate, Codacy, HoundCI, etc) as a PR check? It might be a good way of ensuring we don't have regressions on the code style.

grzuy:
Thanks for the feedback!

We do check for code style regressions on every PR already. TravisCI fails if rubocop fails since https://github.com/kickstarter/rack-attack/pull/351/commits/1e9d601483791b90d8ef6a3c21b14c8bb31eda1c. :-)

PR #416, opened 2019-05-23
Update README example with new ActiveSupport Notification syntax

lmansur:
#405 changed the subscription syntax to comply with `ActiveSupport` specification.

This Pull Request updates an example in the README that used the old syntax.


grzuy:
Awesome, thank you!

PR #419, opened 2019-06-18
Feature proposal: Request instead of Env in callbacks

NikolayRys:
Hello, I would like to propose an improvement that our company uses internally as a monkey-patch, to which I've added the tests, backward compatibility and proper styling for community use.

In short, when a request is throttled or blocklisted, it's much more convenient to have access to a complete Rack request object instead of just environment hash - for example, for the reporting purposes.

This request contains the following:
1. Adds new variables for the callbacks with better naming, which now receive Rack request instead of an environment object.
2. It retains backward compatibility with the previous implementation.
3. Updates to documentation to match the proposed interface.
4. Small improvement for development docs.

Please let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,18 +313,18 @@ Note that `Rack::Attack.cache` is only used for throttling, allow2ban and fail2b
 Customize the response of blocklisted and throttled requests using an object that adheres to the [Rack app interface](http://www.rubydoc.info/github/rack/rack/file/SPEC).
 
 ```ruby
-Rack::Attack.blocklisted_response = lambda do |env|
+Rack::Attack.blocklisted_callback = lambda do |request|
   # Using 503 because it may make attacker think that they have successfully
   # DOSed the site. Rack::Attack returns 403 for blocklists by default
   [ 503, {}, ['Blocked']]
 end
 
-Rack::Attack.throttled_response = lambda do |env|
+Rack::Attack.throttled_callback = lambda do |request|
   # NB: you have access to the name and other data about the matched throttle
-  #  env['rack.attack.matched'],
-  #  env['rack.attack.match_type'],
-  #  env['rack.attack.match_data'],
-  #  env['rack.attack.match_discriminator']
+  #  request.env['rack.attack.matched'],
+  #  request.env['rack.attack.match_type'],
+  #  request.env['rack.attack.match_data'],
+  #  request.env['rack.attack.match_discriminator']
 
   # Using 503 because it may make attacker think that they have successfully
   # DOSed the site. Rack::Attack returns 429 for throttling by default
@@ -411,9 +411,10 @@ def call(env)
   if safelisted?(req)
     @app.call(env)
   elsif blocklisted?(req)
-    self.class.blocklisted_response.call(env)
+    self.class.blocklisted_callback.call(req)
   elsif throttled?(req)
     self.class.throttled_response.call(env)
+    self.class.throttled_callback.call(req)
   else
     tracked?(req)
     @app.call(env)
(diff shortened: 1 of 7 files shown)


NikolayRys:
I would appreciate some feedback about it :) 

To add to my proposal - Rack::Request is much more suited to use in general that Rack environment object: https://www.rubydoc.info/github/rack/rack/master/Rack/Request

grzuy:
Hi @NikolayRys,

Sorry for the delay, and thank you very much for the proposal, extra care to update docs and keep it backwards-compatible.

My thoughts are that if we were to build this from scratch, I would probably agree that yielding the request may be better than the env.

However, I am not sure about making the change now and pushing for a deprecation.

Users that want this feature could still have it by doing `request = Rack::Attack::Request.new(env)` as the first statement of the callback. Given that, I am unsure the total benefit this provide is more the cost of forcing everyone else to update their implementations when we remove the deprecation in the future.

NikolayRys:
Thanks for your feedback. 
Generally, I would agree, but instantiating Rack::Attack::Request feels hackish and is not clearly pointed out in the readme. It looks like an internal class, that potentially might have side-effects during the instantiation. Also the idea to have more than a single request object in the application at one point in time does not occur naturally. 

I agree with you, this may be a rather small improvement, that potentially is a breaking change, but I would suggest to include this in a next major release of your gem - 7.0.0 and do not ship this change alone.

grzuy:
> Thanks for your feedback.
> Generally, I would agree, but instantiating Rack::Attack::Request feels hackish and is not clearly pointed out in the readme. It looks like an internal class, that potentially might have side-effects during the instantiation. Also the idea to have more than a single request object in the application at one point in time does not occur naturally.

Yes, I agree that could feel be a bit hacky.

> 
> I agree with you, this may be a rather small improvement, that potentially is a breaking change, but I would suggest to include this in a next major release of your gem - 7.0.0 and do not ship this change alone.

Worth considering it for 7.0.

Thank you again! :-)

NikolayRys:
UPD: Resolved the conflicts with the current master.

NikolayRys:
UPD: Resolved more conflicts.

NikolayRys:
Okay, there is more conflicts now :)
Quick question, @grzuy, maybe we can merge it at least into a development branch?

NikolayRys:
I've resolved the new conflicts. 

@grzuy, I have a suggestion. I've commented out the deprecation warning, so the people who use the gem won't notice anything if it gets merged right now, but it would allow to use the new callback style too in new projects.
It still contains all other internal changes, including updates for documentation.

When the 7.0 appears on the roadmap, the warnings can be easily uncommented.

NikolayRys:
I've also added tests to cover both styles of callbacks - the new one and the old one.

grzuy:
> I've resolved the new conflicts.
> 
> @grzuy, I have a suggestion. I've commented out the deprecation warning, so the people who use the gem won't notice anything if it gets merged right now, but it will allow to use the new callback style in new projects.

Nice! Good call.

> It still contains all other internal changes, including updates for documentation.
> 
> When the 7.0 appears on the roadmap, the warnings can be easily uncommented.

somanwal:
Is this released? I don't see it in the latest version 6.3.1

grzuy:
> Is this released? I don't see it in the latest version 6.3.1

Not yet.

grzuy:
Finally released this as part of `v6.6.0` with a late minute change to the name of the methods. I found `responder` slightly more precise term than `callback` for these.

Thanks again @NikolayRys!

NikolayRys:
My pleasure 👍

PR #421, opened 2019-07-07
Failsafe on Redis error replies in RedisCacheStoreProxy.

cristiangreco:
RedisCacheStoreProxy will blow up when RedisCacheStore raises a
CommandError exception. In fact, by default the proxied store only
handles BaseConnectionError exceptions, but will let bubble up any other
type of exception from the underlying client.

This pull request uses the same approach from RedisProxy, where store
operations are wrapped in a `rescuing` block that rescues and ignores
BaseError exceptions (the most generic exception class that can be
raised by the Redis client).

Best viewed [ignoring whitespaces](https://github.com/kickstarter/rack-attack/pull/421/files?w=1).

### DIFF
--- a/lib/rack/attack/store_proxy/redis_cache_store_proxy.rb
+++ b/lib/rack/attack/store_proxy/redis_cache_store_proxy.rb
@@ -15,17 +15,33 @@ module Rack
           #
           # So in order to workaround this we use RedisCacheStore#write (which sets expiration) to initialize
           # the counter. After that we continue using the original RedisCacheStore#increment.
-          if options[:expires_in] && !read(name)
-            write(name, amount, options)
+          rescuing do
+            if options[:expires_in] && !read(name)
+              write(name, amount, options)
 
-            amount
-          else
-            super
+              amount
+            else
+              super
+            end
           end
         end
 
+        def read(*_args)
+          rescuing { super }
+        end
+
         def write(name, value, options = {})
-          super(name, value, options.merge!(raw: true))
+          rescuing do
+            super(name, value, options.merge!(raw: true))
+          end
+        end
+
+        private
+
+        def rescuing
+          yield
+        rescue Redis::BaseError
+          nil
         end
       end
     end


cristiangreco:
@grzuy thanks for reviewing this PR and sorry for coming back to you so late!

> To be consistent I guess you would also want to rescue #read, right?

Not sure what you exactly mean here? `read` is not reimplemented by RedisCacheStoreProxy, it's only used within `increment` which is already wrapped in `rescuing` block itself. Or did I misunderstood your suggestion?

grzuy:
Hi @cristiangreco,

Sorry for my short and vague comment :-P

So, rack-attack will eventually call `#read`, `#write` and `#increment` on whichever cache store you configure.

Whether or not those methods are re-implemented on any particular proxy doesn't mean they aren't going to be called.

In particular, for the `RedisCacheStoreProxy`, `#read` is not re-implemented because it wasn't necessary up until now, but it is still being called and delegated to `ActiveSupport::Cache::RedisCacheStore#read`.

If you don't re-implement an rescue `#read` you will continue to get `Redis::BaseError` for those `#read` calls.

cristiangreco:
@grzuy thanks for the thorough explanation! I've force pushed a change to wrap `read` operations as well.

lib/rack/attack/store_proxy/redis_cache_store_proxy.rb
```diff
@@ -15,17 +15,33 @@ def increment(name, amount = 1, options = {})
...
-            super
+              amount
+            else
+              super
+            end
           end
         end
 
+        def read(name)
+          rescuing { super(name) }
```
grzuy: I think we are overriding [`ActiveSupport::Cache::Store#read(name, options = nil)`](https://github.com/rails/rails/blob/b9ca94caea2ca6a6cc09abaffaad67b447134079/activesupport/lib/active_support/cache.rb#L340), which potentially can receive two arguments.

What about going with the following?

```ruby
def read(*_args)
  rescuing { super }
end
```

So that we allow `options` argument, plus we make it forward compatible with any arity changes to the original `#read`?

cristiangreco: That looks sensible, I've updated it!

grzuy:
Thank you!

PR #422, opened 2019-07-10
feat: provide discriminator in throttle_data

### LINKED ISSUE #420: Could we add the discriminator to the throttle_data?
The `throttle_data` stored in `request.env['rack.attack.throttle_data']` currently consists of:
```ruby
        data = {
          count: count,
          period: current_period,
          limit: current_limit,
          epoch_time: epoch_time
        }
```
https://github.com/kickstarter/rack-attack/blob/master/lib/rack/attack/throttle.rb#L33-L38

It would be helpful for me if the discriminator for the current request were added as well:
```ruby
        data = {
          count: count,
          period: current_period,
          limit: current_limit,
          epoch_time: epoch_time,
          discriminator: discriminator
        }

        (request.env['rack.attack.throttle_data'] ||= {})[name] = data
```

I figured that no one else needed this data, or I am thinking about this logic so incorrectly, that what I'm asking for doesn't work. Would it be feasible to add this?

grzuy:
Closes #420

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,7 +354,7 @@ end
 For responses that did not exceed a throttle limit, Rack::Attack annotates the env with match data:
 
 ```ruby
-request.env['rack.attack.throttle_data'][name] # => { :count => n, :period => p, :limit => l, :epoch_time => t }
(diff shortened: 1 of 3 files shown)


NikolayRys:
LTGM, we need it too 👍

Taytay:
That was fast! Thank you!

PR #425, opened 2019-07-27
Example Configuration: black/white -> block/safe

joemasilotti:


grzuy:
Thank you @joemasilotti!

PR #431, opened 2019-08-31
Auto plug middleware for simpler installation

### LINKED ISSUE #429: Auto plug middleware for simpler installation
So we can remove this step https://github.com/kickstarter/rack-attack#plugging-into-the-application, at least for Rails apps can be achieved by defining a rack-attack railtie.

https://api.rubyonrails.org/classes/Rails/Railtie.html

fatkodima:
Closes #429 and several related issues.

It is hard to test that middleware is actually included by default without generating dummy apps inside spec folder for every tested rails version. 
@grzuy maybe you have some ideas how to test this?


fatkodima:
Updated. I dunno how i tested that before locally, but initial implementation was not fully working. We can not use `after_initialize` to setup middleware because at the moment when that callback is called, middleware stack is already built and frozen. So we can not add our own.

There is a solution: to use `initializer`. But we can not use simple `initializer` block (without `before`, `after` parts) because, from [docs](https://guides.rubyonrails.org/configuring.html#using-initializer-files)
> There is no guarantee that your initializers will run after all the gem initializers, so any initialization code that depends on a given gem having been initialized should go into a config.after_initialize block.

And also there is no way to check if middleware was previously included/removed without using rails' private api, so implementation become a little more complicated than I was expected.

Unfortunately, for some unknown reason, can't make new tests pass for rails 4.2, while everything works as expected while manually testing in a local app.

grzuy:
> Unfortunately, for some unknown reason, can't make new tests pass for rails 4.2, while everything works as expected while manually testing in a local app.

I suspect this change (https://github.com/rails/rails/commit/d951558f78cd09074abae4a49e57603b7424ad70) has something to do with the difference between rails 4.2 and 5.0+.

grzuy:
Thanks for updating and adding tests!

fatkodima:
Maybe, not sure. It's weird to me, that in brand new rails 4.2 app while testing manually in console -  everything works, but can't make tests pass. For other versions everything works here and there. I think, this is safe to merge as is.

grzuy:
I understand that it works on Rails 4.2 for you by testing manually via console as of today, but by not having test coverage we are prone to introducing regressions in any future update or refactor.

So I think to be safe we'll want either a working test case or to not make the auto-plug for rails 4.2.

fatkodima:
@grzuy Correct, you are right. Updated to not auto-plug it for rails < 5.

grzuy:
Nice!

Thank you very much for the PR and all the updates!

fatkodima:
@grzuy You probably should configure github's merge button to merge with squash/rebase to make history cleaner for future debuggings.

Issue #427, opened 2019-08-01
Refactor Throttle#matched_by? method


grzuy:
If possible fix CodeClimate issue associated with it. See https://codeclimate.com/github/kickstarter/rack-attack/issues.

grzuy:
Fixed by #435

PR #435, opened 2019-10-06
Refactor `Throttle#matched_by?` method

dsantosmerino:
Hopefully, if it fixes the #427, and sets the number of issues in Code Climate to 0 💪!

As @grzuy said in the mentioned issue, Code Climate complains about the complexity of this method. Here we try to reduce it by using private methods that encapsulate some details that are not required to understand the implementation of the main method.

### DIFF
--- a/lib/rack/attack/throttle.rb
+++ b/lib/rack/attack/throttle.rb
@@ -23,33 +23,49 @@ module Rack
 
       def matched_by?(request)
         discriminator = block.call(request)
+
         return false unless discriminator
 
-        current_period = period.respond_to?(:call) ? period.call(request) : period
-        current_limit  = limit.respond_to?(:call) ? limit.call(request) : limit
-        key            = "#{name}:#{discriminator}"
-        count          = cache.count(key, current_period)
-        epoch_time     = cache.last_epoch_time
+        current_period  = period_for(request)
+        current_limit   = limit_for(request)
+        count           = cache.count("#{name}:#{discriminator}", current_period)
 
         data = {
           discriminator: discriminator,
           count: count,
           period: current_period,
           limit: current_limit,
-          epoch_time: epoch_time
+          epoch_time: cache.last_epoch_time
         }
 
-        (request.env['rack.attack.throttle_data'] ||= {})[name] = data
-
         (count > current_limit).tap do |throttled|
+          annotate_request_with_throttle_data(request, data)
           if throttled
-            request.env['rack.attack.matched']             = name
-            request.env['rack.attack.match_discriminator'] = discriminator
-            request.env['rack.attack.match_type']          = type
-            request.env['rack.attack.match_data']          = data
+            annotate_request_with_matched_data(request, data)
             Rack::Attack.instrument(request)
           end
         end
+      end
+
+      private
+
+      def period_for(request)
+        period.respond_to?(:call) ? period.call(request) : period
+      end
+
+      def limit_for(request)
+        limit.respond_to?(:call) ? limit.call(request) : limit
+      end
+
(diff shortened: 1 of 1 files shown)


lib/rack/attack/throttle.rb
```diff
@@ -23,34 +23,48 @@ def cache
...
+      def annotate_request_with_throttle_data(request, data)
+        (request.env['rack.attack.throttle_data'] ||= {})[name] = data
+      end
+
+      def annotate_request_with_matched_data(request, data)
+        request.env['rack.attack.matched']             = name
+        request.env['rack.attack.match_discriminator'] = data[:discriminator]
+        request.env['rack.attack.match_type']          = type
+        request.env['rack.attack.match_data']          = data
+        Rack::Attack.instrument(request)
```
grzuy: I feel it might be a bit misleading or unexpected that a method named "annotate request with something" ends up sending a notification with the `instrument` method here.

What about leaving it outside?

lib/rack/attack/throttle.rb
```diff
@@ -23,34 +23,48 @@ def cache
...
+
         return false unless discriminator
 
-        current_period = period.respond_to?(:call) ? period.call(request) : period
-        current_limit  = limit.respond_to?(:call) ? limit.call(request) : limit
-        key            = "#{name}:#{discriminator}"
-        count          = cache.count(key, current_period)
-        epoch_time     = cache.last_epoch_time
+        current_period  = current_period(request)
+        current_limit   = current_limit(request)
```
grzuy: Can we please use different names for local variable and instance methods? I think it helps avoid possible confusion for the developer :-)

Some ideas:
- `period_for(request)` and `limit_for(request)` or
-  `request_period(request)` and `request_limit(request)`

fatkodima:
Personally, original code was more clear than a bunch of extracted one liners, as I have all in one place and don't need to jump back and forth to understand what's happening here. So I would left it as it was. Not always warning is an indication of a problem and sometimes it is better to ignore. 

What should definitely (maybe not completely) improve code climate reporting is changing following lines:
```ruby
(request.env['rack.attack.throttle_data'] ||= {})[name] = data
```
To
```ruby
request.env['rack.attack.throttle_data'] ||= {}
request.env['rack.attack.throttle_data'][name] = data
```
But again, original line is shorter and cleaner, as for me.

And lines
```ruby
(count > current_limit).tap do |throttled|
  if throttled
    ...
```

To
```ruby
if count > current_limit # throttled?
  ...
```

lib/rack/attack/throttle.rb
dsantosmerino: You're right, changed!

lib/rack/attack/throttle.rb
dsantosmerino: Agree, could be misleading. Fixed!

dsantosmerino:
Thanks for your feedback, @fatkodima. In my opinion, you don't need to jump to all the private methods, unless you need to know the implementation detail. The method names should be descriptive enough to understand what is going on in the main `#matched_by?` method. However, I see your point, and we don't need to please always the linter.

It's my first attempt to contribute here, so of course, you know more about the code style of this gem and if it this change could be positive or negative. Feel free to decide whatever you want, no problem 😃

grzuy:
While I think there's probably room for more refactoring here, I do think this is a step forward.

Thank you @dsantosmerino!

grzuy:
>  Not always warning is an indication of a problem and sometimes it is better to ignore.

@fatkodima I agree with you about this statement, though I also think there's an improvement by encapsulating the setting of request env data.

Thank you for sharing your perspective.

Issue #301, opened 2018-03-15
Better integration with user app tests


grzuy:
It happened, at least to some users, that it was not clear how to integrate with their app tests:

https://github.com/kickstarter/rack-attack/issues/249
https://github.com/kickstarter/rack-attack/issues/228
https://github.com/kickstarter/rack-attack/issues/220

Feels like there's an opportunity for improvement here.

Maybe it's just mentioning something in the README about how users are supposed to deal with rack-attack in their test suites and/or providing a new feature that let's the user configure rack-attack in test-mode or similar. Not sure yet.

---
Sub tasks:

* [x] Improve test case isolation (issue: https://github.com/kickstarter/rack-attack/issues/249)
* [x] Selectively disable Rack::Attack for some tests (issue: #384)
* [x] Document new features in "Testing" README section

grzuy:
Also related: https://github.com/kickstarter/rack-attack/pull/218

grzuy:
Related: https://github.com/rubygems/rubygems.org/pull/1414#discussion_r77869334

grzuy:
Added [Testing](https://github.com/kickstarter/rack-attack#testing) section to the README.

Issue #249, opened 2017-11-08
Resetting throttle count between tests


theirishpenguin:
Just wondering, what is the best way to reset the rack attack counters between "it" blocks in a spec, so that throttling can be tested independently in different "it" blocks.

I presume one would use an around block like this...

```
        around(:each) do |each|
            clear_rack_attack_cache
            each.run
            clear_rack_attack_cache
        end
```

Then perhaps...

```
        def clear_rack_attack_cache
            while Rack::Attack.cache.store.instance_variable_get(:@data).keys.length > 0
                Rack::Attack.cache.store.clear
                sleep(1)
            end
        end
```
FYI, using ActiveSupport::Cache::MemoryStore as the backing store. Is a sleep required or is this synchronous?

Either way, it doesn't seem to be working quite right for me as I don't always get the expected number of 200's before I get a 429.

markedmondson:
I'm doing this, it seems to work...

```ruby
    Rack::Attack.throttles.each do |k, throttle|
      Rack::Attack.cache.reset_count("#{k}:#{key}", throttle.period)
    end
```

grzuy:
For the record,

> It appears to me that Rack::Attack.clear! is a bit confusing in that is resetting some things, not everything.

`clear!` has been renamed to `clear_configuration` to avoid further confusion in https://github.com/kickstarter/rack-attack/pull/344.

> think this is something we should add explicit support for, so users don't need to hack the cache store in order to achieve test case isolation in their app tests.

Issue with this goal is the following https://github.com/kickstarter/rack-attack/issues/301

grzuy:
We should consider adding a `Rack::Attack.reset!` method, which should reset all `Rack::Attack`-specific state (mostly throttles, fail2ban and allow2ban counters) so that users can for example do something like:

```rb
config.before do
  Rack::Attack.reset!
end
```

when using RSpec.

grzuy:
> We should consider adding a `Rack::Attack.reset!` method, which should reset all `Rack::Attack`-specific state (mostly throttles, fail2ban and allow2ban counters) so that users can for example do something like:
> 
> ```ruby
> config.before do
>   Rack::Attack.reset!
> end
> ```
> 
> when using RSpec.

`Rack::Attack.reset!` has been implemented and merged in https://github.com/kickstarter/rack-attack/pull/436.

PR #436, opened 2019-10-10
Allow to reset state

fatkodima:
https://github.com/kickstarter/rack-attack/issues/249#issuecomment-539029878

I can think of 2 (not functionally equal) implementations:
1. `Rack::Attack.reset!` which really resets (deletes) specific keys in cache store and can be used for testing as well as production. But, while this can be implemented for most of cache stores (using https://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html#method-i-delete_matched), this is not (and can't be) supported by memcache (the second most popular cache store), because to reset state, we should know exact key names, which is impossible for production considering rack-attack's dynamic key names nature. Unfortunately, memcache does not support deleting based on patterns/prefixes, like in redis. And need to figure out how to use this with custom, created by user, cache stores.
Also, personally, I don't think this is a very popular usecase to reset *production* rack-attack cache. I have used this once to shrink redis memory usage, implemented as a simple rake task executing `scan` and `del` on redis. I think, for general use, changing keys prefix and let the backed store prune old entries here is a valid solution.   

2. Implementation only to support user tests. I implemented this version here. Example of usage:
```ruby
# test_helper.rb
Rack::Attack.test_mode
```

```ruby
# some_test_file.rb
...
setup do
  Rack::Attack.cache.store.clear
end
...
```

After agreeing on implementation, will add missing tests and documentation.

### DIFF
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -14,6 +14,7 @@ module Rack
     class Error < StandardError; end
     class MisconfiguredStoreError < Error; end
     class MissingStoreError < Error; end
+    class IncompatibleStoreError < Error; end
 
     autoload :Check,                'rack/attack/check'
     autoload :Throttle,             'rack/attack/throttle'
@@ -51,6 +52,10 @@ module Rack
       def clear!
         warn "[DEPRECATION] Rack::Attack.clear! is deprecated. Please use Rack::Attack.clear_configuration instead"
         @configuration.clear_configuration
+      end
+
+      def reset!
+        cache.reset!
       end
 
       extend Forwardable
--- a/lib/rack/attack/cache.rb
+++ b/lib/rack/attack/cache.rb
@@ -41,6 +41,17 @@ module Rack
         store.delete("#{prefix}:#{unprefixed_key}")
       end
 
+      def reset!
+        if store.respond_to?(:delete_matched)
+          store.delete_matched("#{prefix}*")
+        else
+          raise(
+            Rack::Attack::IncompatibleStoreError,
+            "Configured store #{store.class.name} doesn't respond to #delete_matched method"
+          )
+        end
+      end
+
       private
 
       def key_and_expiry(unprefixed_key, period)
--- a/lib/rack/attack/store_proxy/redis_proxy.rb
+++ b/lib/rack/attack/store_proxy/redis_proxy.rb
@@ -43,6 +43,19 @@ module Rack
           rescuing { del(key) }
         end
 
+        def delete_matched(matcher, _options = nil)
+          cursor = "0"
+
+          rescuing do
+            # Fetch keys in batches using SCAN to avoid blocking the Redis server.
+            loop do
+              cursor, keys = scan(cursor, match: matcher, count: 1000)
+              del(*keys) unless keys.empty?
+              break if cursor == "0"
+            end
+          end
+        end
+
         private
 
         def rescuing
--- a/spec/rack_attack_spec.rb
+++ b/spec/rack_attack_spec.rb
@@ -99,4 +99,26 @@ describe 'Rack::Attack' do
       end
     end
   end
+
+  describe 'reset!' do
+    it 'raises an error when is not supported by cache store' do
+      Rack::Attack.cache.store = Class.new
+      assert_raises(Rack::Attack::IncompatibleStoreError) do
+        Rack::Attack.reset!
+      end
+    end
+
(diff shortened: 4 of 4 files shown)


grzuy:
> [#249 (comment)](https://github.com/kickstarter/rack-attack/issues/249#issuecomment-539029878)
> 
> I can think of 2 (not functionally equal) implementations:
> 
>     1. `Rack::Attack.reset!` which really resets (deletes) specific keys in cache store and can be used for testing as well as production. But, while this can be implemented for most of cache stores (using https://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html#method-i-delete_matched), this is not (and can't be) supported by memcache (the second most popular cache store), because to reset state, we should know exact key names, which is impossible for production considering rack-attack's dynamic key names nature. Unfortunately, memcache does not support deleting based on patterns/prefixes, like in redis. And need to figure out how to use this with custom, created by user, cache stores.
>        Also, personally, I don't think this is a very popular usecase to reset _production_ rack-attack cache. I have used this once to shrink redis memory usage, implemented as a simple rake task executing `scan` and `del` on redis. I think, for general use, changing keys prefix and let the backed store prune old entries here is a valid solution.
> 

Good call.

I would still consider providing a `Rack::Attack.reset!` method that depends on the store `#delete_matched` method (passing the prefix) which will work for redis and memory_store.

For mem_cache users wanting to reset in their test suites will have the option to:
1. use the memory_store for the test suite and use this new `.reset!` method
2. clear all mem_cache during tests


>     2. Implementation only to support user tests. I implemented this version here. Example of usage:
> 
> 
> ```ruby
> # test_helper.rb
> Rack::Attack.test_mode
> ```
> 
> ```ruby
> # some_test_file.rb
> ...
> setup do
>   Rack::Attack.cache.store.clear
> end
> ...
> ```
> 
> After agreeing on implementation, will add missing tests and documentation.

fatkodima:
@grzuy Updated with your suggestions. I haven't added any readme notes. I think it is better to add all of them as third subtask of #301

lib/rack/attack.rb
```diff
@@ -53,6 +53,19 @@ def clear!
...
 
+      def reset!
+        store = cache.store
+
+        if store.respond_to?(:delete_matched)
+          store.delete_matched("#{cache.prefix}*")
+        else
+          raise(
+            MisconfiguredStoreError,
+            "Configured store #{store.class.name} doesn't respond to #delete_matched method"
```
grzuy: `MisconfiguredStoreError` was conceived to let a user know it need to reconfigure rack-attack or nothing will work.

In this particular case, not sure we're exactly there. This would more like "Everything will work fine with your current store but you won't be able to use the `#reset!` method"? Maybe `IncompatibleStoreError` for the constant?

lib/rack/attack.rb
```diff
@@ -53,6 +53,19 @@ def clear!
         @configuration.clear_configuration
       end
 
+      def reset!
+        store = cache.store
+
+        if store.respond_to?(:delete_matched)
+          store.delete_matched("#{cache.prefix}*")
+        else
```
grzuy: Sounds like everything in this method implementation is interacting with `cache`-only stuff (`store` and `prefix`).
Should we move `reset!` inside `Cache` and make the implementation here just `cache.reset!`?

fatkodima:
Updated.

grzuy:
Awesome, thank you!

camilova:
Should `Rack::Attack::reset!` remove current IPs that I have listed with `safelist_ip` ?
Because I'm trying to remove the IPs, but the new `reset!` method doesnt remove them and does not raise any error. Is there a way to remove from cache only one specific IP that I was added through `safelist_ip`?

jdelStrother:
Any chance of getting a new release with this in?

grzuy:
Released in [`v6.3.0`](https://github.com/kickstarter/rack-attack/blob/master/CHANGELOG.md#630---2020-04-26).

grzuy:
> Should `Rack::Attack::reset!` remove current IPs that I have listed with `safelist_ip` ?
> Because I'm trying to remove the IPs, but the new `reset!` method doesnt remove them and does not raise any error. Is there a way to remove from cache only one specific IP that I was added through `safelist_ip`?

Hi @camilova , `safelist_ip` doesn't write anything to cache.

camilova:
Thanks! I had a misconception, but now I know that `.clear_configuration`
is what I was looking for.

PR #438, opened 2019-10-11
Make discriminators case-insensitive by default

### LINKED ISSUE #395: Mixed-case value for discriminator problem
I've recently implemented throttling with Rack::Attack and figured that the login throttling could be circumvented by switching the case of a case-insensitive email. I resolved the problem by returning a downcased value from the discriminator block.

I believe Rack::Attack should treat discriminator as a case-insensitive value by default, while providing an option to disable the behavior conditionally.

Any thoughts on this? I'd be willing to submit a PR with the change.

fatkodima:
Closes #395 

I think, as also mentioned in original issue, discriminators should not only be `downcase`d, but also `strip`ed and `chomp`ed to avoid fooling counters by adding whitespaces.
I'm open for better name to `case_insensitive_discriminator` setting.

### DIFF
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -31,7 +31,7 @@ module Rack
     autoload :Allow2Ban,            'rack/attack/allow2ban'
 
     class << self
-      attr_accessor :enabled, :notifier
+      attr_accessor :enabled, :notifier, :discriminator_normalizer
       attr_reader :configuration
 
       def instrument(request)
@@ -79,6 +79,9 @@ module Rack
     # Set defaults
     @enabled = true
     @notifier = ActiveSupport::Notifications if defined?(ActiveSupport::Notifications)
+    @discriminator_normalizer = lambda do |discriminator|
+      discriminator.to_s.strip.downcase
+    end
     @configuration = Configuration.new
 
     attr_reader :configuration
--- a/lib/rack/attack/throttle.rb
+++ b/lib/rack/attack/throttle.rb
@@ -22,8 +22,7 @@ module Rack
       end
 
       def matched_by?(request)
-        discriminator = block.call(request)
-
+        discriminator = discriminator_for(request)
         return false unless discriminator
 
         current_period  = period_for(request)
@@ -48,6 +47,14 @@ module Rack
       end
 
       private
+
+      def discriminator_for(request)
+        discriminator = block.call(request)
+        if discriminator && Rack::Attack.discriminator_normalizer
+          discriminator = Rack::Attack.discriminator_normalizer.call(discriminator)
+        end
+        discriminator
+      end
 
       def period_for(request)
         period.respond_to?(:call) ? period.call(request) : period
--- a/spec/rack_attack_throttle_spec.rb
+++ b/spec/rack_attack_throttle_spec.rb
@@ -144,3 +144,47 @@ describe 'Rack::Attack.throttle with block retuning nil' do
     end
   end
 end
+
+describe 'Rack::Attack.throttle with discriminator_normalizer' do
+  before do
+    @period = 60
+    @emails = [
+      "person@example.com",
+      "PERSON@example.com ",
+      " person@example.com\r\n  ",
+    ]
+    Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
+    Rack::Attack.throttle('logins/email', limit: 4, period: @period) do |req|
+      if req.path == '/login' && req.post?
+        req.params['email']
+      end
+    end
+  end
+
+  it 'should not differentiate requests when discriminator_normalizer is enabled' do
+    post_logins
+    key = "rack::attack:#{Time.now.to_i / @period}:logins/email:person@example.com"
+    _(Rack::Attack.cache.store.read(key)).must_equal 3
+  end
+
+  it 'should differentiate requests when discriminator_normalizer is disabled' do
+    begin
+      prev = Rack::Attack.discriminator_normalizer
+      Rack::Attack.discriminator_normalizer = nil
+
+      post_logins
+      @emails.each do |email|
+        key = "rack::attack:#{Time.now.to_i / @period}:logins/email:#{email}"
+        _(Rack::Attack.cache.store.read(key)).must_equal 1
+      end
+    ensure
+      Rack::Attack.discriminator_normalizer = prev
+    end
+  end
+
+  def post_logins
+    @emails.each do |email|
+      post '/login', email: email
+    end
+  end
+end


eliotsykes:
> I'm open for better name to case_insensitive_discriminator setting.

To allow for more control, how about a `discriminator_normalizer` or `discriminator_sanitizer` config that is a callable, with a default, e.g.

```rb
config.discriminator_normalizer = ->(discriminator) { discriminator.to_s.downcase.gsub(/\s+/, "") }
```

Consider stripping whitespaces from *inside* the discriminator, and not just surrounding whitespace to prevent more throttle bypasses.

This would mitigate when authentication code normalizes the email before lookup, including removing inner spaces so `"foo@example.com"` and `"f o o @example.com"` all allow successful login but bypass rack attack's throttle. The clearance gem does this, there may be others.

fatkodima:
@eliotsykes Lovely 👍 Updated with your suggestion.

spec/rack_attack_throttle_spec.rb
```diff
@@ -148,3 +148,38 @@
...
+    Rack::Attack.throttle('logins/email', limit: 2, period: @period) do |req|
+      if req.path == '/login' && req.post?
+        req.params['email']
+      end
+    end
+  end
+
+  it 'should not differentiate requests when discriminator_normalizer is enabled' do
+    post '/login', email: 'person@example.com'
+    post '/login', email: 'PERSON@example.com'
```
eliotsykes: Consider adding a few more email variants that will give confidence surrounding and inner, single+multiple consecutive whitespaces are normalized.

lib/rack/attack.rb
```diff
@@ -146,6 +146,9 @@ def clear!
       retry_after = (env['rack.attack.match_data'] || {})[:period]
       [429, { 'Content-Type' => 'text/plain', 'Retry-After' => retry_after.to_s }, ["Retry later\n"]]
     end
+    @discriminator_normalizer = lambda do |discriminator|
+      discriminator.to_s.downcase.chomp.gsub(/\s+/, "")
```
eliotsykes: Consider checking if the `.chomp` call is still needed, it may be covered by the `gsub` call.

lib/rack/attack/throttle.rb
```diff
@@ -51,6 +51,16 @@ def matched_by?(request)
           end
         end
       end
+
+      private
+
+      def discriminator_for(request)
+        discriminator = block.call(request)
+        if discriminator && Rack::Attack.discriminator_normalizer != false
```
eliotsykes: Consider dropping the ` != false` on end of line 59, so this will gracefully handle when discriminator_normalizer is set to `nil` or `false`.

spec/rack_attack_throttle_spec.rb
```diff
@@ -148,3 +148,38 @@
     end
   end
 end
+
+describe 'Rack::Attack.throttle with case sensivity option for discriminator' do
```
eliotsykes: Consider updating description to be more generalized, omitting the case-sensitivity text.

lib/rack/attack.rb
fatkodima: Still needed, because it removes `\r` and `\n` while regex removes only all kinds of whitespaces

fatkodima:
Updated.

lib/rack/attack.rb
eliotsykes: Try this in `irb`:

```
irb(main):014:0> "hello\r\n\r\r\n".strip
=> "hello"
irb(main):015:0> "hello\r\n\r\r\n".chomp
=> "hello\r\n\r"
irb(main):016:0> "hello\r\n\r\r\n".gsub(/\s+/, "")
=> "hello"
irb(main):017:0> 
```

fatkodima: Ah, my bad. I used single quotes instead of double quotes for string. Nice catch! Updated.

grzuy:
@eliotsykes Thank you very much for helping with the code review!

> > I'm open for better name to case_insensitive_discriminator setting.
> 
> To allow for more control, how about a `discriminator_normalizer` or `discriminator_sanitizer` config that is a callable, with a default, e.g.
> 
> ```ruby
> config.discriminator_normalizer =
>   ->(discriminator) { discriminator.to_s.downcase.gsub(/\s+/, "") }
> ```
> 
> Consider stripping whitespaces from _inside_ the discriminator, and not just surrounding whitespace to prevent more throttle bypasses.
> 
> This would mitigate when authentication code normalizes the email before lookup, including removing inner spaces so `"foo@example.com"` and `"f o o @example.com"` all allow successful login but bypass rack attack's throttle. The clearance gem is susceptible to this, there may be others.

I am quite comfortable downcasing and removing leading and trailing whitespace, but I feel that we're doing a bit "too much" if we remove inside spaces.

I understand the `f o o @example.com` email use case, but I am not quite convinced this is something Rack::Attack should be trying to normalize. It sounds to me that's probably something that would end up being an invalid email for any authentication algorithm.

We can reconsider if we find there's several other uses cases for this, but I'd rather start small with this normalization :-)

fatkodima:
@grzuy Updated with your suggestion.

eliotsykes:
> I understand the f o o @example.com email use case, but I am not quite convinced this is something Rack::Attack should be trying to normalize. It sounds to me that's probably something that would end up being an invalid email for any authentication algorithm.

That's what I used to think. Unfortunately it is the default behaviour of the `clearance` gem's authentication algorithm, it normalizes the email before looking it up in the db.

clearance's normalization scrubs all whitespace, including inner whitespace (it also downcases), see:

https://github.com/thoughtbot/clearance/blob/7d6def116148d2cdd389c7c05e9119d86e6a1d3a/lib/clearance/user.rb#L115-L129

lib/rack/attack.rb
```diff
@@ -79,6 +79,9 @@ def clear!
     # Set defaults
     @enabled = true
     @notifier = ActiveSupport::Notifications if defined?(ActiveSupport::Notifications)
+    @discriminator_normalizer = lambda do |discriminator|
+      discriminator.to_s.strip.chomp.downcase
```
eliotsykes: I suspect the `chomp` call is not needed thanks to the `strip` call.

grzuy:
> > I understand the f o o @example.com email use case, but I am not quite convinced this is something Rack::Attack should be trying to normalize. It sounds to me that's probably something that would end up being an invalid email for any authentication algorithm.
> 
> That's what I used to think. Unfortunately it is the default behaviour of the `clearance` gem's authentication algorithm, it normalizes the email before looking it up in the db.
> 
> During authentication, clearance's normalization scrubs all whitespace, including inner whitespace (it also downcases), see:
> 
> https://github.com/thoughtbot/clearance/blob/7d6def116148d2cdd389c7c05e9119d86e6a1d3a/lib/clearance/user.rb#L115-L129

@eliotsykes Gotcha.

Thanks for sharing, that's interesting to see.

Still feel it would be "too much" to remove inside spaces in Rack::Attack's discriminator default normalization, which we cannot assume is always an email.

I prefer to be on the safe side for a default.

Anyone can perform extra normalization if necessary for their particular use case, which in the case of clearance users throttling by email would be `User.normalize_email(discriminator)` I guess.

grzuy:
Thank you @fatkodima for this and all your other contributions!

fatkodima:
@grzuy Agree with you for "too much". 
"f o o @example.com" is an invalid email (99% sure), so it is strange to me that clearance accepts that.

eliotsykes:
> Still feel it would be "too much" to remove inside spaces in Rack::Attack's discriminator default normalization, which we cannot assume is always an email.
> I prefer to be on the safe side for a default.

Cool, sounds like the right choice and very reasonable 👍

grzuy:
Released in `v6.5.0`.
Was renamed to `Rack::Attack.throttle_discriminator_normalizer`.

PR #440, opened 2019-10-11
Allow to configure Retry-After header for default throttled_response handler

### LINKED ISSUE #209: throttled requests seem to have a fixed retry after value independently of the time left to expire

fatkodima:
Closes #209

### DIFF
--- a/lib/rack/attack/configuration.rb
+++ b/lib/rack/attack/configuration.rb
@@ -4,7 +4,7 @@ module Rack
   class Attack
     class Configuration
       attr_reader :safelists, :blocklists, :throttles, :anonymous_blocklists, :anonymous_safelists
-      attr_accessor :blocklisted_response, :throttled_response
+      attr_accessor :blocklisted_response, :throttled_response, :throttled_response_retry_after_header
 
       def initialize
         @safelists = {}
@@ -13,11 +13,18 @@ module Rack
         @tracks = {}
         @anonymous_blocklists = []
         @anonymous_safelists = []
+        @throttled_response_retry_after_header = false
 
         @blocklisted_response = lambda { |_env| [403, { 'Content-Type' => 'text/plain' }, ["Forbidden\n"]] }
         @throttled_response   = lambda do |env|
-          retry_after = (env['rack.attack.match_data'] || {})[:period]
-          [429, { 'Content-Type' => 'text/plain', 'Retry-After' => retry_after.to_s }, ["Retry later\n"]]
+          if throttled_response_retry_after_header
+            match_data = env['rack.attack.match_data']
+            now = match_data[:epoch_time]
+            retry_after = match_data[:period] - (now % match_data[:period])
+            [429, { 'Content-Type' => 'text/plain', 'Retry-After' => retry_after.to_s }, ["Retry later\n"]]
+          else
+            [429, { 'Content-Type' => 'text/plain' }, ["Retry later\n"]]
+          end
         end
       end
 
@@ -86,6 +93,7 @@ module Rack
         @tracks = {}
         @anonymous_blocklists = []
         @anonymous_safelists = []
+        @throttled_response_retry_after_header = false
       end
     end
   end
--- a/spec/acceptance/throttling_spec.rb
+++ b/spec/acceptance/throttling_spec.rb
@@ -20,7 +20,7 @@ describe "#throttle" do
     get "/", {}, "REMOTE_ADDR" => "1.2.3.4"
 
     assert_equal 429, last_response.status
-    assert_equal "60", last_response.headers["Retry-After"]
+    assert_nil last_response.headers["Retry-After"]
     assert_equal "Retry later\n", last_response.body
 
     get "/", {}, "REMOTE_ADDR" => "5.6.7.8"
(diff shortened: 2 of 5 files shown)


spec/acceptance/throttling_spec.rb
```diff
@@ -34,6 +34,27 @@
     end
   end
 
+  it "returns correct Retry-After header if enabled" do
+    begin
+      prev = Rack::Attack.retry_after_header_enabled
+      Rack::Attack.retry_after_header_enabled = true
```
grzuy: I know I suggested this name in https://github.com/kickstarter/rack-attack/issues/209#issuecomment-398841599, but giving it a second thought, should we name it `Rack::Attack.throttled_response_retry_after_header = true` so that it's clear to the user that the setting is scoped to the throttled response only and not something else?

spec/acceptance/throttling_spec.rb
```diff
@@ -34,6 +34,27 @@
...
+      Rack::Attack.throttle("by ip", limit: 1, period: 60) do |request|
+        request.ip
+      end
+
+      get "/", {}, "REMOTE_ADDR" => "1.2.3.4"
+      assert_equal 200, last_response.status
+
+      Timecop.travel(25) do
+        get "/", {}, "REMOTE_ADDR" => "1.2.3.4"
+        assert last_response.headers["Retry-After"].to_i <= 60 - 25
```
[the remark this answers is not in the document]
fatkodima: As far as I understand rack-attack internals, this can not not be done exactly due to how expiration time is calculated https://github.com/kickstarter/rack-attack/blob/9cf227a32ec1b3e122166f00ee72e6570f305c5d/lib/rack/attack/cache.rb#L46-L51
as this depends on which exact second/minute this is calculated.
Or am I missing something and then how this can be done?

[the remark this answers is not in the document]
grzuy: Yeah, tricky.

But I think it can be done by making the test run aligned with epoch = 0.

```diff
diff --git a/spec/acceptance/throttling_spec.rb b/spec/acceptance/throttling_spec.rb
index a9021c6..20b5882 100644
--- a/spec/acceptance/throttling_spec.rb
+++ b/spec/acceptance/throttling_spec.rb
@@ -43,13 +43,17 @@ describe "#throttle" do
         request.ip
       end
 
-      get "/", {}, "REMOTE_ADDR" => "1.2.3.4"
+      Timecop.freeze(Time.at(0)) do
+        get "/", {}, "REMOTE_ADDR" => "1.2.3.4"
+      end
+
       assert_equal 200, last_response.status
 
-      Timecop.travel(25) do
+      Timecop.freeze(Time.at(25)) do
         get "/", {}, "REMOTE_ADDR" => "1.2.3.4"
-        assert last_response.headers["Retry-After"].to_i <= 60 - 25
       end
+
+      assert_equal "35", last_response.headers["Retry-After"]
     ensure
       Rack::Attack.throttled_response_retry_after_header = prev
     end
```

spec/acceptance/throttling_spec.rb
```diff
@@ -34,6 +34,27 @@
...
+      end
+    ensure
+      Rack::Attack.throttled_response_retry_after_header = prev
```
grzuy: I think we would not need this if this new config setting is cleared as part of `clear_configuration`.

spec/acceptance/throttling_spec.rb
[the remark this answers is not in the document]
fatkodima: Cleverly 👍

fatkodima:
Updated.

grzuy:
Thank you very much!

grzuy:
Released in [`v6.3.0`](https://github.com/kickstarter/rack-attack/blob/master/CHANGELOG.md#630---2020-04-26).

lib/rack/attack/configuration.rb
```diff
@@ -13,11 +13,18 @@ def initialize
...
+        @throttled_response_retry_after_header = false
 
         @blocklisted_response = lambda { |_env| [403, { 'Content-Type' => 'text/plain' }, ["Forbidden\n"]] }
         @throttled_response   = lambda do |env|
-          retry_after = (env['rack.attack.match_data'] || {})[:period]
-          [429, { 'Content-Type' => 'text/plain', 'Retry-After' => retry_after.to_s }, ["Retry later\n"]]
+          if throttled_response_retry_after_header
+            match_data = env['rack.attack.match_data']
+            now = match_data[:epoch_time]
+            retry_after = match_data[:period] - (now % match_data[:period])
```
sonalkr132, line 23: hi, don't we need `+ 1` here to match [expired_in](https://github.com/kickstarter/rack-attack/blob/master/lib/rack/attack/cache.rb#L65)?

grzuy: Seems like we do...
Thanks @sonalkr132 

@fatkodima Any thoughts about this?

fatkodima: I can't say for 100% sure right now, but seems like we don't need that +1.
In `expired_in` that +1 is used to set key expiration key, but here `retry_after` describes when the next period will begin to start accepting requests.
@sonalkr132 can you please investigate more, if it looks like I'm not correct.

sonalkr132: yes, you are right. acceptance of new requests depends on the period window and not key expiry time. Thank you for explaining,

PR #442, opened 2019-10-13
Add ability to use multiple instances of middleware

### LINKED ISSUE #178: RFC: Issues with multiple Attack instances in the same process / Rack stack (class ivar binding)
I would like to use multiple instances of Rack::Attack in the same Rack stack, but in different places. Each one of these would accept it's own throttles, blacklists etc. Judging from the code, I assumed I could do

     class MyThrottle < Rack::Attack
        throttle(...) do |req|
          ...
        end
     end

It turned out, however, that the method performing the throttling still reaches into the attribute reader of the superclass, and thus finds throttles defined on the parent instance. I think the same is the case for blacklisting and whitelisting, but I don't necessarily need them in this specific application. I was only able to make throttles work in a subclass by doing

     class MyThrottle < Rack::Attack
        ...
        def throttled?(req)
          self.class.throttles.any? do |name, throttle|
            throttle[req]
          end
        end
     end

This stems from the fact that some smart machinery is involved in making the configuration available on the class level. I think that it might make more sense if the same configuration could be applied with, say, options (since block handling with middleware attachment is not that well understood):

     my_attack_config = Rack::Attack::Config.new do |c|
        c.throttle(...) do |req| # etc
          ...
        end
     end
     use Rack::Attack, config: my_attack_config

So I think there was some thought given to separating the various classes' configurations within one Ruby process, but this was not really tested. You could in theory alter the `throttled?` etc. methods to call into the right class for the variables and consider this a bug, but maybe the API can just be improved and handle this explicitly?

fatkodima:
This can be used as:
```ruby
use Rack::Attack do
  ...
  throttle("requests by ip", limit: 5, period: 2) do |request|
    request.ip
  end

  blocklist("block all access to admin") do |request|
    request.path.start_with?("/admin")
  end

  self.blocklisted_response = lambda do |env|
    [503, {}, ['Blocked']]
  end
  ...
end
```

The code is a little bit more complicated than it should be as I'm not broke old style of reopening `Rack::Attack`. I would prefer to remain only new style of configuring, but then this will force all gem users to update their `Rack::Attack` settings and recently added feature of using middleware automatically would not make sense. Maybe this makes more sense for 7.0 release?

New tests and documentation are missing - would add them after agreeing on implementation. 

Closes #178

### DIFF
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -84,9 +84,15 @@ module Rack
 
     attr_reader :configuration
 
-    def initialize(app)
+    def initialize(app, &block)
       @app = app
-      @configuration = self.class.configuration
+      @configuration =
+        if block_given?
+          configuration = Configuration.new
+          configuration.instance_exec(&block)
+        else
+          self.class.configuration
+        end
     end
 
     def call(env)


grzuy:
> would add them after agreeing on implementation.

Hi @fatkodima ,

I think I see an opportunity to split this PR in 2 and have separate discussions.

If I understood correctly, I see that if you leave the `initialize`/`instance_exec` part out of this PR, it would only be a refactor encapsulating the config in a `Configuration` object, which I think it is great and provides a lot of value for making code clear and having separation of concerns. That is something I would be ready to merge.

Then we can discuss the other piece which is the one actually making user-facing changes.

fatkodima:
Removed code for extracted Configuration class.

lib/rack/attack.rb
```diff
@@ -84,9 +84,15 @@ def clear!
...
     attr_reader :configuration
 
-    def initialize(app)
+    def initialize(app, &block)
       @app = app
-      @configuration = self.class.configuration
+      @configuration =
+        if block_given?
+          configuration = Configuration.new
+          configuration.instance_exec(&block)
```
jellybob, line 92: You need to make the final statement in this branch `configuration`, otherwise `@configuration` gets set to the return value of the last statement in the block, which in my case was an instance of `Throttle`.

Otherwise, this seems to work nicely, I've monkey patched this change into a project I'm working on it does the job wonderfully.

ioquatix:

grzuy (commented):
Thanks @fatkodima ,

Happy to merge this if we add some test cases using the added optional block argument

lib/rack/attack.rb
grzuy: Thanks @jellybob for the code review.

Ideally this problem would reveal in any new test case covering the new block argument.

fatkodima:
Updated.

spec/acceptance/rails_middleware_spec.rb
```diff
@@ -16,5 +16,17 @@
       @app.initialize!
       assert @app.middleware.include?(Rack::Attack)
     end
+
+    it "can be configured via a block" do
```
santib, line 20: the PR goal (per its title) seems to be supporting the ability to have multiples instances of Rack::Attack. I'm wondering.. should we have a test that demonstrates how that works? Not only for catching regressions but also to document this new "feature". Thoughts?

santib:
Hi, I think this is a cool feature to add 💯 

Probably it's minor, but wanted to leave here my only concern that is that if someone in the future uses the gem like:
```ruby
use(Rack::Attack) do
  blocklist("my blocklist") do |req|
    req.path == '/login'
  end
end
```

and then runs `Rack::Attack.blocklists` (or any other method on the class), it will return the value from the class config which will differ. I mean, it makes sense, it's just that it can confuse end users while debugging, maybe? Not a blocker to go ahead.

franzliedke:
@fatkodima Great suggestion! Can I help out with the missing test? (I was working on a similar PR until @santib pointed me here...)

spec/acceptance/rails_middleware_spec.rb
santib, line 20: I just tried [if this test file was even running](https://github.com/fatkodima/rack-attack/compare/multiple-instances...santib:multiple-instances-santib?expand=1) and the [CI passed](https://github.com/santib/rack-attack/actions/runs/7476191579)..

I'm working on a fix

santib: https://github.com/rack/rack-attack/pull/649

santib: when I tried rebasing on that branch so that this test file is run, the test fails

```
$ BUNDLE_GEMFILE=gemfiles/rails_7_1.gemfile be ruby -Itest spec/acceptance/rails_middleware_spec.rb

.F

Failure:
Middleware for Rails#test_0002_can be configured via a block [spec/acceptance/rails_middleware_spec.rb:26]:
Expected: 403
  Actual: 200

bin/rails test spec/acceptance/rails_middleware_spec.rb:20
```

santib:
> Would appreciate 🙏 if someone would help me fix CI for older rubies.

Seems like there are failures in every Ruby version (not just old ones) e.g. [Ruby 3.3](https://github.com/rack/rack-attack/actions/runs/8951229444/job/24587320296?pr=442).

Note: I updated your branch with `main` since it was missing some fixes on the CI.

fatkodima:
Fixed tests.

spec/acceptance/middleware_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    app.initialize!
+    assert app.middleware.include?(Rack::Attack)
+  end
+
+  def app
+    Rack::Builder.new do
+      use Rack::Attack do
+        blocklist_ip("1.2.3.4")
+      end
```
santib, line 23: can we actually test that multiple instances of the middleware work together?

Maybe I'm missing something but I tried

```ruby
  def app
    Rack::Builder.new do
      use Rack::Attack do
        blocklist_ip("1.2.3.4")
      end
      use Rack::Attack do
        blocklist_ip("4.3.2.1")
      end

      run lambda { |_env| [200, {}, ['Hello World']] }
    end.to_app
  end

  it "can be configured via a block" do
    get "/", {}, "REMOTE_ADDR" => "1.2.3.4"
    assert_equal 403, last_response.status

    get "/", {}, "REMOTE_ADDR" => "4.3.2.1"
    assert_equal 403, last_response.status
  end
```

and it didn't work 🤔 thoughts?

fatkodima: Multiple instances on the same rack app are not working because of this guard clause https://github.com/rack/rack-attack/blob/b6ce502b1491ed4de057d1dc20b8391f8d7b3d19/lib/rack/attack.rb#L105

santib: ahh ok I see.. that was introduced to [guarantee idempotency](https://github.com/rack/rack-attack/pull/457). But with this PR, that's no longer accurate.

What I mean is that in the past it was safe to just have `Rack::Attack` execute once because all configs were the same, but with this PR we could have different configs of `Rack::Attack` and only the first one to run will execute, the rest will silently be skipped, right?

At this point I'm starting to doubt if the benefits of the PR are greater than its drawbacks, especially around generating confusion to end users and inflicting pain on them 🤔

Pros I see:
- Nicer way to config using the block rather than the `Rack::Attack` class.

Cons I see:
- There are now 2 ways to config `Rack::Attack` which can increase confusion, especially since you can't use both at the same time.
- Idempotency guard no longer works as expected, now only the config of the first middleware to run, will take place.
- Potential conflicts between the automatically added middleware to Rails apps and the usage of the middleware with the new block method.

WDYT? Please let me know your thoughts, I'm open to discuss it and change my mind 🙌

ioquatix: I think this is still a good direction.

Ideally we retain compatibility with any existing configuration mechanism and redirect it to a per-instance configuration.

I don't think the idempotency fix is correct. It seems like XY problem.

fatkodima: @ioquatix Do you have suggestions how to proceed?

I wish the middleware was not automatically added to rails apps some time ago 😐.

ioquatix: I don't have time to provide detailed feedback until after RubyKaigi, but I can take a look then. Is it part of Rails' default stack?

fatkodima: > Is it part of Rails' default stack?

Yes, https://github.com/rack/rack-attack/blob/b6ce502b1491ed4de057d1dc20b8391f8d7b3d19/lib/rack/attack/railtie.rb#L11-L15

> but I can take a look then

👍

santib: Maybe we can introduce some configs so that you either get the old way to config + idempotency + the railtie, OR you get the new way to config rack attack with the block + ability to have multiple middlewares added + no railtie?

What I mean is that I’m ok with both approaches, but I think the key here is not mixing both of them at the same time 🤔

ioquatix: It's fairly straight forward to configure using a Railtie, so we could have a default middleware which gets configured via Rails' own configuration system, but that doesn't mean we can't have non-global instances.

PR #443, opened 2019-10-14
Add test coverage

fatkodima:
I need [Test Coverage Id](https://docs.codeclimate.com/docs/finding-your-test-coverage-token) from owners to finish this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ bin
 .ruby-version
 .ruby-gemset
 .byebug_history
+coverage/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,8 @@
+env:
+  global:
+    - CC_TEST_REPORTER_ID=MISSING_ID_HERE
+    - COVERAGE=true
+
 dist: xenial
 language: ruby
 cache: bundler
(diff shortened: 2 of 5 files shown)


grzuy:
Hi @fatkodima,

I extremely value test suite **quality**. I really do. A lot. Which was the main driver for https://github.com/kickstarter/rack-attack/pull/269 and https://github.com/kickstarter/rack-attack/issues/293.

That said, I am not sure "Test Coverage" percentage indicators are very good at indicating high or low test suite quality. Also, I am worried it can lead people to over test with low-value/low-quality test cases just for the sake of getting a higher number. In general, I tend to agree with almost everything that Martin Fowler wrote in https://martinfowler.com/bliki/TestCoverage.html.

I can see the value of making it easy for anyone to run simplecov on demand so that someone can report an issue about any seriously untested code path, but not sure about promoting this indicator as a key metric for the quality of the project.

Thank you for the PR and your concern of improving the test suite :-)

fatkodima:
Hi, @grzuy 
Yes, I'm mostly sharing your opinions. For me, only really low test coverage percentage is meaningful, as it indicates about very untested code. But high percentage is a bit meaningless due to the reasons mentioned in your comment.

This metric will only indicate that this project is enough tested and should not motivate people to improve test coverage by squeezing additional percent fractions. Feel free to close it if you have doubts.

Would you mind reviewing other PRs? :)

PR #444, opened 2019-10-14
Use double quotes in code

fatkodima:
Mixed style is used everywhere, so it is time to use only one.


grzuy:
This was intentionally not set.

While I value enforcing conventions for the sake of readability and consistency, I don't feel like mixing single and double quotes makes an significant impact in readability to justify enforcing a style an incur in the cost of introducing extra friction to regular developers and new contributors.

PR #445, opened 2019-10-14
Fix rescuing errors in RedisProxy#increment

### LINKED ISSUE #437: `RedisProxy` will still raise an error when Redis is down
As far as I see, the idea behind [rescuing proxy errors](https://github.com/kickstarter/rack-attack/blob/95347e35f10a145c2ab09d8a799f4702cf490c87/lib/rack/attack/store_proxy/redis_proxy.rb#L52-L56) is to not disrupt the application when there is a problem with Redis. However, if the Redis connection goes away, invocations inside `Redis#pipelined` will return an instance of `Redis::Future` before raising `Redis::CannotConnectError`, hence we'll still get an error (Redis::FutureNotReady) when [resolving its value](https://github.com/kickstarter/rack-attack/blob/95347e35f10a145c2ab09d8a799f4702cf490c87/lib/rack/attack/store_proxy/redis_proxy.rb#L43)

rack-attack version: 6.1.0
Redis gem version: 4.1.2
Rails version: 5.2.3

Can you please confirm that this is not intended?

fatkodima:
Closes #437

### DIFF
--- a/lib/rack/attack/store_proxy/redis_proxy.rb
+++ b/lib/rack/attack/store_proxy/redis_proxy.rb
@@ -31,16 +31,12 @@ module Rack
         end
 
         def increment(key, amount, options = {})
-          count = nil
-
           rescuing do
             pipelined do
-              count = incrby(key, amount)
+              incrby(key, amount)
               expire(key, options[:expires_in]) if options[:expires_in]
-            end
+            end.first
           end
-
-          count.value if count
         end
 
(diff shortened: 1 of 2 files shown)


fatkodima:
Updated.

grzuy:
Triggered rebuild of the failing job. Green now.

grzuy:
Awesome!
Thank you very much.

PR #446, opened 2019-10-16
Extract Configuration class

fatkodima:
Extraction for #442

### DIFF
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -14,6 +14,7 @@ module Rack
     class MisconfiguredStoreError < Error; end
     class MissingStoreError < Error; end
 
+    autoload :Configuration,        'rack/attack/configuration'
     autoload :Cache,                'rack/attack/cache'
     autoload :Check,                'rack/attack/check'
     autoload :Throttle,             'rack/attack/throttle'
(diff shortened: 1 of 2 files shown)


lib/rack/attack.rb
```diff
@@ -31,82 +32,8 @@ class MissingStoreError < Error; end
...
-        end
-      end
-
-      def tracked?(request)
-        tracks.each_value do |track|
-          track.matched_by?(request)
-        end
-      end
+      attr_accessor :enabled, :notifier
+      attr_reader :config
```
grzuy: Can we please use non abbreviated name, i.e. `configuration`?

lib/rack/attack.rb
```diff
@@ -158,19 +76,16 @@ def call(env)
       env['PATH_INFO'] = PathNormalizer.normalize_path(env['PATH_INFO'])
       request = Rack::Attack::Request.new(env)
 
-      if safelisted?(request)
+      if @config.safelisted?(request)
```
grzuy: Can we please use the reader instead of the instance variable so that it's easier to detect typos?

fatkodima:
Updated.

grzuy:
:clap:

PR #447, opened 2019-10-16
Speedup CI by running all appraisals in one travis job

fatkodima:
Currently ci is running 1 hour 15 minutes in average. This should *significantly* improve this time. Let's see, if I'm correct.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,30 +13,17 @@ before_install:
   - gem update --system
   - gem install bundler -v "~> 2.0"
 
-gemfile:
-  - gemfiles/rack_2_0.gemfile
-  - gemfiles/rack_1_6.gemfile
-  - gemfiles/rails_6_0.gemfile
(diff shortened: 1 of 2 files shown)


fatkodima:
Ok, speedup is not *as* significant, because I was fooled by travis' ci namings - 1 hour 15 min is **sum** across all jobs, while actual run time is near 16 minutes on average.
This change reduces runtime to less than 5 minutes (> 3 times speedup). 
@grzuy Wdyt, is this change worth it?

grzuy:
> @grzuy Wdyt, is this change worth it?

I think it doesn't.
At least, as of today, don't see the TravisCI build time as something we need to fix.

Thank you for taking a stab at this anyway!
Might be helpful in the future.

fatkodima:
Yes, if it was really 1.15 hour, then we would definitely need to fix that 😀

PR #450, opened 2019-10-17
Do not rescue all errors for redis backed stores

fatkodima:
It is a very bad idea to rescue **all** redis errors (including errors for invalid commands). We can't know when we messed things up in code, because all errors will be silently swallowed. It is better to rescue only connection-related errors. `RedisCacheStore` already rescues connection errors, so no need to do it manually.

### DIFF
--- a/lib/rack/attack/store_proxy/redis_cache_store_proxy.rb
+++ b/lib/rack/attack/store_proxy/redis_cache_store_proxy.rb
@@ -15,33 +15,17 @@ module Rack
           #
           # So in order to workaround this we use RedisCacheStore#write (which sets expiration) to initialize
           # the counter. After that we continue using the original RedisCacheStore#increment.
-          rescuing do
-            if options[:expires_in] && !read(name)
-              write(name, amount, options)
+          if options[:expires_in] && !read(name)
+            write(name, amount, options)
 
-              amount
-            else
-              super
-            end
+            amount
+          else
+            super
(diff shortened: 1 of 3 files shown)


grzuy:
Nice find. I agree.

Thank you!

grzuy:
Released in `v6.3.0`.

cristiangreco:
Hey @fatkodima @grzuy 👋 

This PR re-introduces a problem which was fixed in https://github.com/kickstarter/rack-attack/pull/421

How's one supposed to handle a `CommandError` exception? For example, if the connection is ok but Redis is busy loading the database into memory, it'll return a `-LOADING` response which will raise a `CommandError` in the application. What is the best way to handle such errors when they bubble up?

grzuy:
> Hey @fatkodima @grzuy wave
> 
> This PR re-introduces a problem which was fixed in #421
> 
> How's one supposed to handle a `CommandError` exception? For example, if the connection is ok but Redis is busy loading the database into memory, it'll return a `-LOADING` response which will raise a `CommandError` in the application. What is the best way to handle such errors when they bubble up?

Hi @cristiangreco,

Yes, we had some back and forth with that. Sorry for any inconvenience.

Can you please open an issue reporting what see it's happening in the latest version vs what you expect to happen so that we can discuss there, others chime in, and have a decision on what we want to do?

Thanks in advance!

PR #451, opened 2019-10-20
Reduce redis roundtrips when incrementing counters

fatkodima:
From 2 to 1, for each throttle.

### DIFF
--- a/lib/rack/attack/store_proxy/active_support_redis_store_proxy.rb
+++ b/lib/rack/attack/store_proxy/active_support_redis_store_proxy.rb
@@ -13,16 +13,13 @@ module Rack
         end
 
         def increment(name, amount = 1, options = {})
-          # #increment ignores options[:expires_in].
-          #
-          # So in order to workaround this we use #write (which sets expiration) to initialize
-          # the counter. After that we continue using the original #increment.
-          if options[:expires_in] && !read(name)
-            write(name, amount, options)
-
-            amount
-          else
-            super
+          rescuing do
+            with do |conn|
+              conn.multi do
+                conn.incrby(name, amount)
+                conn.expire(name, options[:expires_in]) if options[:expires_in]
+              end.first
+            end
           end
         end
 
@@ -32,6 +29,14 @@ module Rack
 
         def write(name, value, options = {})
           super(name, value, options.merge!(raw: true))
+        end
+
+        private
+
+        def rescuing
+          yield
+        rescue Redis::BaseConnectionError
+          nil
         end
       end
     end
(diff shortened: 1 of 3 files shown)


lib/rack/attack/store_proxy/redis_cache_store_proxy.rb
```diff
@@ -11,22 +11,27 @@ def self.handle?(store)
...
-            amount
-          else
-            super
+          rescuing do
+            redis.with do |conn|
+              conn.multi do
+                conn.incrby(name, amount)
+                conn.expire(name, options[:expires_in]) if options[:expires_in]
+              end.first
+            end
```
grzuy: This was intentionally done this way in https://github.com/kickstarter/rack-attack/commit/3af7394b6ac6ae97e2cd63eda18daa92a29d0e6b.

My goal at that moment was, and still is, to have Rack::Attack know as less as possible about cache backend details.

lib/rack/attack/store_proxy/active_support_redis_store_proxy.rb
```diff
@@ -13,16 +13,13 @@ def self.handle?(store)
...
-            amount
-          else
-            super
+          rescuing do
+            with do |conn|
+              conn.multi do
+                conn.incrby(name, amount)
+                conn.expire(name, options[:expires_in]) if options[:expires_in]
+              end.first
+            end
```
grzuy: See description in https://github.com/kickstarter/rack-attack/pull/385 when this one done.

> [..] don't want to depend on any internals of externally developed cache store. We only wrap/proxy/adapt to the public interface using delegation.

lib/rack/attack/store_proxy/redis_cache_store_proxy.rb
grzuy: There's one optimization which can be done without re-implementing which is only `read`ing for activesupport < `v6.0`, because expiration support was added in `v6.0` https://github.com/rails/rails/pull/33254/commits/9d5b02ec5062a23665ec596ef7d3efe4f5abcc27.

lib/rack/attack/store_proxy/active_support_redis_store_proxy.rb
grzuy: Similar comment here: can be optimized by not `read`ing for `redis-activesupport` >= `v5.2`, which added expiration support for `#increment` https://github.com/redis-store/redis-activesupport/commit/a794a315d658d8daae40370a23e6240a46471ee1.

fatkodima:
Ah, haven't seen that. I got your point. I will rewrite this then using mentioned optimizations.

fatkodima:
@grzuy Ok, so I have checked `rails`s and `redis-activesupport`s implementations - theirs `increment` method is doing 2 commands (not sure why they don't use pipelines/transactions) to redis, so won't be any performance improvements. I'm closing this. 
Thanks for review! 😀

PR #453, opened 2019-10-24
Make store proxies lookup dynamic

fatkodima:
https://github.com/kickstarter/rack-attack/pull/441#discussion_r338270775


lib/rack/attack/store_proxy.rb
```diff
@@ -1,20 +1,26 @@
...
-        MemCacheStoreProxy,
-        RedisStoreProxy,
-        RedisProxy,
-        RedisCacheStoreProxy,
-        ActiveSupportRedisStoreProxy
-      ].freeze
+    class StoreProxy < SimpleDelegator
+      class << self
+        def proxies
+          @@proxies ||= []
```
grzuy: Would instance variable be enough here?

fatkodima: No, should be class variable to work with chains of inheritance, like `class RedisProxy < StoreProxy`, `class RedisStoreProxy < RedisProxy`

grzuy: Gotcha.

Was thinking under the assumption all proxies inherited directly from `StoreProxy`, forgot that chained redis case.

spec/rack_attack_dalli_proxy_spec.rb
```diff
@@ -2,9 +2,9 @@
 
 require_relative 'spec_helper'
 
-describe Rack::Attack::StoreProxy::DalliProxy do
+describe Rack::Attack::StoreProxies::DalliProxy do
   it 'should stub Dalli::Client#with on older clients' do
-    proxy = Rack::Attack::StoreProxy::DalliProxy.new(Class.new)
+    proxy = Rack::Attack::StoreProxies::DalliProxy.new(Class.new)
```
grzuy: I know this is not "officially" part of the public API, but given it might be easy to avoid in this case, can we please try not breaking any user code that references the current proxies, by keeping constant refs the same as before?

Maybe the `StoreProxy` class you added can become `BaseProxy` or just `Base`?

fatkodima: Yes, sure this can be make. But if we will consider original part of this extracted pr, 1) this namespace will still be changed to something like `StoreAdapters` and 2) personally, new names (`StoreProxies` namespace and `StoreProxy` base class) looks cleaner to me than `Base` or `BaseProxy`.
 
So, taking into account first mentioned reason, should I still preserve original namespace name?

fatkodima:
Ignore my last comment.
Updated with your suggestion. I will change names in original pr then.

lib/rack/attack/store_proxy/redis_store_proxy.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'delegate'
+require 'rack/attack/store_proxy/redis_proxy'
```
grzuy: Nice find

grzuy:
:clap:

grzuy:
Released in `v6.5.0`.
Thanks again.

PR #455, opened 2019-10-28
Fix RedisStoreProxy write when passing expires_in option

dgmora:
Version: HEAD
Redis verison: 4.1.3
Rails version: 5.1.7

This happened while throttling requests. While I was using rack-attack 6.0 I was running into exceptions when redis was offline and noticed #445, so I updated rack-attack.  After that, I got a different exception:

```ruby
> wrong number of arguments (given 4, expected 3)

redis (4.1.3) lib/redis.rb, line 819
------------------------------------

  814     #
  815     # @param [String] key
  816     # @param [Fixnum] ttl
  817     # @param [String] value
  818     # @return [String] `"OK"`
> 819     def setex(key, ttl, value)
  820       synchronize do |client|
  821         client.call([:setex, key, ttl, value.to_s])
  822       end
  823     end
  824   

Full backtrace
--------------
 - redis (4.1.3) lib/redis.rb:819:in `setex'
 - /usr/local/lib/ruby/2.5.0/delegate.rb:83:in `method_missing'
 -  () usr/local/bundle/bundler/gems/rack-attack-a7ce9a89fe68/lib/rack/attack/store_proxy/redis_store_proxy.rb:19:in `block in write'
 -  () usr/local/bundle/bundler/gems/rack-attack-a7ce9a89fe68/lib/rack/attack/store_proxy/redis_proxy.rb:62:in `rescuing'
 -  () usr/local/bundle/bundler/gems/rack-attack-a7ce9a89fe68/lib/rack/attack/store_proxy/redis_store_proxy.rb:19:in `write'
 -  () usr/local/bundle/bundler/gems/rack-attack-a7ce9a89fe68/lib/rack/attack/cache.rb:79:in `do_count'
 -  () usr/local/bundle/bundler/gems/rack-attack-a7ce9a89fe68/lib/rack/attack/cache.rb:26:in `count'
 -  () usr/local/bundle/bundler/gems/rack-attack-a7ce9a89fe68/lib/rack/attack/throttle.rb:30:in `matched_by?'
 -  () usr/local/bundle/bundler/gems/rack-attack-a7ce9a89fe68/lib/rack/attack/configuration.rb:77:in `block in throttled?'
 -  () usr/local/bundle/bundler/gems/rack-attack-a7ce9a89fe68/lib/rack/attack/configuration.rb:76:in `any?'
 -  () usr/local/bundle/bundler/gems/rack-attack-a7ce9a89fe68/lib/rack/attack/configuration.rb:76:in `throttled?'
 -  () usr/local/bundle/bundler/gems/rack-attack-a7ce9a89fe68/lib/rack/attack.rb:108:in `call'
```

Looking at the [docs](https://www.rubydoc.info/github/redis/redis-rb/Redis#setex-instance_method), it doesn't seem possible to pass an `options` arguments there.

I did not add any tests since this there weren't any for `lib/rack/attack/store_proxy/redis_store_proxy.rb` and this seemed pretty specific, so it shouldn't go to the offline examples.

After doing so, rack attack continued working correctly.

### DIFF
--- a/lib/rack/attack/store_proxy/redis_store_proxy.rb
+++ b/lib/rack/attack/store_proxy/redis_store_proxy.rb
@@ -16,7 +16,7 @@ module Rack
 
         def write(key, value, options = {})
           if (expires_in = options[:expires_in])
-            rescuing { setex(key, expires_in, value, raw: true) }
+            rescuing { setex(key, expires_in, value) }
           else
             rescuing { set(key, value, raw: true) }
           end


grzuy:
Hi @dgmora,

Thank you for taking the time to debug and contribute.

Would like to to understand better why is that your app is using the `RedisStoreProxy` but it ends up in `lib/redis.rb` `setex` with 4 arguments.

Can you please share your Rack::Attack store backend configuration for your app?
Are you setting `Rack::Attack.cache.store = ...`?

dgmora:
Hi, sorry for the delay. 
`Rack::Attack::StoreProxy::RedisStoreProxy` inherits from `Rack::Attack::StoreProxy::RedisProxy` and `BaseProxy`. I think that since the method has a different signature (3 vs 4 arguments) there's a `method_missing` and then it gets delegated to `redis` because `setex` is not implemented in `RedisProxy` or `BaseProxy` itself.

This is our setup in `config/initializers/rack_attack.rb`:
```ruby  
Rack::Attack.throttle(...)

ActiveSupport::Notifications.subscribe(/rack_attack/) do
  . . .
end

Rack::Attack.cache.store = Rack::Attack::StoreProxy::RedisStoreProxy.new(Redis.new(url: <url>))
```

I have checked the docs, and also `set` does not seem to accept the option `raw: true`, so I am not sure if it's actually achieving anything in the `else` in `RedisStoreProxy`. I don't have the whole picture here, but maybe it'd be possible to remove this whole method from `RedisStoreProxy` and just inherit it from `RedisProxy`?

grzuy:
Hi @dgmora ,

`RedisStoreProxy` is meant to proxy `Redis::Store` and `RedisProxy` is meant to proxy `Redis`, as can be seen in their respective `.handle?` methods.

So in your case you need to change the setup to:

```ruby
Rack::Attack.cache.store = Rack::Attack::StoreProxy::RedisProxy.new(Redis.new(url: <url>))
```

dgmora:
Hi, I'll give this a try. Anyway, isn't the implementation still wrong?

PR #457, opened 2019-10-29
fix: avoid unintended effects on load_config_initializers and other gems load order

### LINKED ISSUE #452: Unexpected interaction with I18n
I'm trying to generate a small test case app to demonstrate this issue, but we experienced something odd when we tried to upgrade our Rails 6.0.0 app from rack-attack 6.1.0 to 6.2.0. The PR with the update, which does nothing beyond bump the one line in Gemfile.lock for rack-attack, results in all of our explicitly defined `time.formats` in our I18n locale files being replaced with the out-of-the-box formats distributed by Rails. We use YAML files in `config/locales/*.yml` to set our formats, and when I inspect them with `I18n.backend.translations[:en][:time]`, I can see that with rack-attack 6.2.0 they are no longer what they normally are in our app.

I'm completely unclear how this could happen, as I can't find references to I18n or `formats` in the gem, but flipping the gem version back and forth in our development and test environments causes the formats in our `config/locales/en.yml` file to be wiped out.

### LINKED ISSUE #456: v6.2.0 incompatible with sequel-rails
v6.1.0 and previous were okay, but v6.2.0 is doing something that is causing my models to be loaded before `sequel-rails` sets up the database connection resulting in 

```
Sequel::Error:
  No database associated with Sequel::Model: have you called Sequel.connect or Sequel::Model.db= ?
```
I've spent the last while trying to figure out how to get things loaded in the same order as before, but I've run into only dead ends. 

It appears the problem happens in both rails 5.2.3 and rails 6.0.0.

Does anyone have any clues?

grzuy:
Because of the sort algorithm rails uses to satisfy `after` and `before`
constraints, gems can have unintended effects on others. See
https://github.com/rails/rails/commit/0a120a818d413c64ff9867125f0b03788fc306f8

Prefer making rack-attack middleware idempotent instead of relying on
the load order and the contents of the middleware stack too much.

closes #452
closes #456

### DIFF
--- a/lib/rack/attack/railtie.rb
+++ b/lib/rack/attack/railtie.rb
@@ -3,17 +3,9 @@
 module Rack
   class Attack
     class Railtie < ::Rails::Railtie
-      initializer 'rack.attack.middleware', after: :load_config_initializers, before: :build_middleware_stack do |app|
+      initializer "rack-attack.middleware" do |app|
         if Gem::Version.new(::Rails::VERSION::STRING) >= Gem::Version.new("5.1")
-          middlewares = app.config.middleware
-          operations = middlewares.send(:operations) + middlewares.send(:delete_operations)
-
-          use_middleware = operations.none? do |operation|
-            middleware = operation[1]
-            middleware.include?(Rack::Attack)
-          end
-
-          middlewares.use(Rack::Attack) if use_middleware
+          app.middleware.use(Rack::Attack)
         end
       end
     end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -46,6 +46,8 @@ class MiniTest::Spec
     Rack::Builder.new do
       # Use Rack::Lint to test that rack-attack is complying with the rack spec
       use Rack::Lint
+      # Intentionally added twice to test idempotence property
+      use Rack::Attack
       use Rack::Attack
       use Rack::Lint
 
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -153,8 +153,9 @@ module Rack
     end
 
     def call(env)
-      return @app.call(env) unless self.class.enabled
+      return @app.call(env) if !self.class.enabled || env["rack.attack.called"]
 
+      env["rack.attack.called"] = true
       env['PATH_INFO'] = PathNormalizer.normalize_path(env['PATH_INFO'])
       request = Rack::Attack::Request.new(env)
 
--- a/spec/acceptance/rails_middleware_spec.rb
+++ b/spec/acceptance/rails_middleware_spec.rb
@@ -18,12 +18,6 @@ if defined?(Rails)
         assert_equal 1, @app.middleware.count(Rack::Attack)
       end
 
-      it "is not added when it was added explicitly" do
-        @app.config.middleware.use(Rack::Attack)
-        @app.initialize!
-        assert_equal 1, @app.middleware.count(Rack::Attack)
-      end
-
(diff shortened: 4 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -46,6 +46,8 @@ def app
     Rack::Builder.new do
       # Use Rack::Lint to test that rack-attack is complying with the rack spec
       use Rack::Lint
+      # Intentionally added twice to test idempotence property
+      use Rack::Attack
```
geoffharcourt: 👍

lib/rack/attack/railtie.rb
```diff
@@ -3,17 +3,9 @@
 module Rack
   class Attack
     class Railtie < ::Rails::Railtie
-      initializer 'rack.attack.middleware', after: :load_config_initializers, before: :build_middleware_stack do |app|
+      initializer "rack-attack.middleware" do |app|
         if Gem::Version.new(::Rails::VERSION::STRING) >= Gem::Version.new("5.1")
```
fatkodima: With new implementation, I think this restriction can be removed.

fatkodima:
Clean and simple 👏 This should be the initial implementation 😀

lib/rack/attack/railtie.rb
grzuy: Probably.

I wan't this change to be the minimal possible update that fixes the issue, which can be released as a patch-level update (`v6.2.1`).

If we open this up for more rails version, that can be part of `v6.3.0`.

Thanks for reviewing.

jarthod:
Hi, I just upgraded from 4.x to the latest version and noticed all these changes about the middleware being inserted by default (which is a good idea I agree). Though I tried moving it because I need it to be before some other middleware of ours (to avoid some database query). I tried `gem 'rack-attack', require: false` and:

```ruby
# config/application.rb
...
require 'rack/attack'
config.middleware.use CompressedRequests
...
```

But the middleware would still end-up after every others:

```
> rake middleware
...
use CompressedRequests
use Rack::Attack
...
```
Because of the initializer hook I guess. So AFAICS I don't have any other option than inserting it twice?
Is this really the preferred/official way to move this middleware now? It does not sound super efficient to have it twice :/ (I understand the re-entrance check is quick but that's still additional code for no value at every request).

I am not requesting a change here (especially if I am the only one who needs this), just asking maybe if there's a workaround to change the placing? or a way to skip the auto-add? (I checked the code to see if I could require other files to skip the railties but it's part of the main file so I can't).

Thanks for your answer.

ps: And thanks for `rack-attack`! it's nice and clean and very useful :bow:

grzuy:
Hi @jarthod , thank you for the thorough explanation of your concern.

So, just doing

```rb
config.middleware.insert_after(Rack::Attack, CompressedRequests)
```

doesn't work, does it?

jarthod:
Indeed it doesn't, because the `Rack::Attack` middleware is not inserted yet, so this (in `application.rb`):
```ruby
    require 'rack/attack'
    config.middleware.insert_after(Rack::Attack, CompressedRequests)
```
Raises: `No such middleware to insert after: Rack::Attack (RuntimeError)`

Also even if it did, I wouldn't be able to change the insertion line for every middleware I use, especially the ones in gems.
For the same reason, I can't remove the middleware manually (with `config.middleware.delete`) before adding it back elsewhere.

Unless maybe I manage to define an initializer which is gonna be executed AFTER the one from rack-attack, but I'm not sure how to guarantee that, it sounds brittle.

PR #460, opened 2019-11-20
Fail2ban earlier return when already banned

styd:
Hi, it would be nice if we can return earlier from fail2ban filter when it's already banned.

### DIFF
--- a/lib/rack/attack/fail2ban.rb
+++ b/lib/rack/attack/fail2ban.rb
@@ -5,16 +5,14 @@ module Rack
     class Fail2Ban
       class << self
         def filter(discriminator, options)
+          # Return true for blocklist
+          return true if banned?(discriminator)
+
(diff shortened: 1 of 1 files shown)


fatkodima:
What problem are you trying to solve?

styd:
> What problem are you trying to solve?

Hi, thanks for commenting on this PR.
I literally have no problem against the current situation. You guys have done a great job.
I was just wondering if we still need to check for the options when the requests match the discriminator has already been banned. It certainly have passed the options checking before it was banned, right? (Please, CMIIW).
So, I just thought it would be nice to skip the options checking if banned.

grzuy:
Hi @styd,

Thank you for your PR!
I think it's cleaner to always check needed arguments first. We don't want to have different "method signature" enforcement based on external state (cache state). I think it may lead users to confusion.

Thanks again :-)

PR #482, opened 2020-05-18
Fix deprecation warning in Rails 5.2.4.3

rofreg:
Rails 5.2.4.3 was released today, with a [security fix](https://github.com/rails/rails/commit/467e3399c9007996c03ffe3212689d48dd25ae99) for [CVE-2020-8165](https://groups.google.com/forum/#!topic/rubyonrails-security/bv6fW4S0Y1c). As a result of this fix, Rack::Attack now throws deprecation warnings, because it writes a value to cache with `raw: true` but reads it back in a subsequent request WITHOUT `raw: true`.

### DIFF
--- a/lib/rack/attack/store_proxy/redis_cache_store_proxy.rb
+++ b/lib/rack/attack/store_proxy/redis_cache_store_proxy.rb
@@ -15,7 +15,7 @@ module Rack
           #
           # So in order to workaround this we use RedisCacheStore#write (which sets expiration) to initialize
           # the counter. After that we continue using the original RedisCacheStore#increment.
-          if options[:expires_in] && !read(name)
+          if options[:expires_in] && !read(name, raw: true)
             write(name, amount, options)
 
             amount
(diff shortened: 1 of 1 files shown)


lib/rack/attack/store_proxy/redis_cache_store_proxy.rb
```diff
@@ -15,7 +15,7 @@ def increment(name, amount = 1, options = {})
           #
           # So in order to workaround this we use RedisCacheStore#write (which sets expiration) to initialize
           # the counter. After that we continue using the original RedisCacheStore#increment.
-          if options[:expires_in] && !read(name)
+          if options[:expires_in] && !read(name, raw: true)
```
grzuy, line 18: This of course fixes this particular `read` invocation, but in order to fix them all I think we need to override `#read` with `raw: true`, similar to what we have for `#write`.

grzuy: ...and that would mean this particular change can be reverted.

rofreg: Ah, thank you! Given that `write` has been overridden so it always uses `raw: true`, makes sense that `read` would need to do the same – and I can see the project is already doing this for `ActiveSupportRedisStoreProxy` 🙂 PR update coming shortly.

grzuy:
Awesome, thank you!

Issue #459, opened 2019-11-08
v6.2 auto-load of middleware creates side-effects


zarqman:
Upon upgrading to 6.2.1 (from 6.1.0), we end up with Rack::Attack twice in our middleware stack. 

Actual behavior:
```
$ rails middleware
...
use Rack::Head
use Rack::ConditionalGet
use Authenticator::PartOne
use Rack::Attack     # <--- our intended Rack::Attack instance
use Authenticator::PartTwo
use Rack::Attack     # <--- the duplicate now added in 6.2 that we don't want
run OurApp::Application.routes
```

Expected behavior:
```
$ rails middleware
...
use Authenticator::PartOne
use Rack::Attack     # <--- our intended Rack::Attack instance
use Authenticator::PartTwo
run OurApp::Application.routes
```

As you can see, our specific case involves a two-phase authentication middleware stack that is specifically designed to surround Rack::Attack, so we cannot move Rack::Attack to the end of the stack, where it installs itself by default.

I see that 6.2.1 was intended to turn the second instance into a no-op, but we'd much prefer avoiding the double load to begin with.

Given the need to configure Rack::Attack anyway, one could argue against auto-loading the middleware. However, I'd like to instead propose a couple of ways to allow disabling the auto-load for advanced configurations like ours, while preserving it as the new default behavior.

1. Add a config option `Rack::Attack.auto_load = true`. Defaults to true. When false, the initializer would skip the auto-load.

2. Rework the gem's default require by moving most of lib/rack/attack.rb into a new file. Replace the original file with nothing but:
```ruby
require 'rack/attack/<newfile>'
require 'rack/attack/railtie' if defined?(::Rails)
```
This would allow advanced configurations to set the Gemfile as
```ruby
gem 'rack-attack', require: 'rack/attack/<newfile>'
```
which would implicitly disable the auto-load railtie.

Either way preserves your new default while adding a path for more specific configuration for those of us with apps with advanced middleware configurations.

Are you open to either of the above solutions? If at least one is acceptable, I'm happy to prepare a PR accordingly if that'd be helpful.

FWIW, either of the above would also allow the removal of the already-called condition (`return ... if ... env["rack.attack.called"]`) added in #457, which I think may break/prevent PR #442.

grzuy:
Hi @zarqman, 

Thank you for the detailed issue report!

How are you inserting `Authenticator::PartTwo` middleware?

zarqman:
[...]

PR #491, opened 2020-07-01
Configure middleware position in Rails

pfac:
Why:

* The position of a middleware in the Rack attack has consequences. For
  instance, any instrumentation middleware like [Skylight] will measure
  and capture what happens in the middlewares below it, but not on those
  above.

  The current implementation forces the rack-attack middleware to always
  be appended to the list when initializing Rails, making it the last on
  the list when added, but often others will be added afterwards.

  While this behaviour is perfectly suitable by default, there are
  situations where one might need to control where the middleware is on
  the list. Given the lack of configuration points to tweak this
  behaviour, currently the only way to achieve this is to add the
  middleware again at the top. This results in a duplication of the
  `Rack::Attack` middleware in the stack.

This change addresses the need by:

* Shamelessly copying the approach followed by [skylight-ruby], where
  the Railtie initializes a configuration object that can be set up in
  `config/application.rb`, thus allowing the position of the middleware
  to be tweaked.

[Skylight]: https://skylight.io
[skylight-ruby]: https://github.com/skylightio/skylight-ruby

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,21 @@ You can disable it permanently (like for specific environment) or temporarily (c
 Rack::Attack.enabled = false
 ```
 
+You can also tweak its position in the Rack middleware stack.
+
+```ruby
+# in config/application.rb
+
+# by index
+config.rack_attack.middleware_position = 0
+
+# relative to another middleware
+config.rack_attack.middleware_position = {
+  before: AnotherMiddleware, # optional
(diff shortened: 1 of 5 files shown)


pfac:
I guess this might be useful to address #459 too.

pfac:
Closing this due to lack of response.

PR #495, opened 2020-08-20
Rescue InvalidParameterError with 400 Response

tubbo:
The documentation for [Rack::QueryParser#parse_nested_query][] states that when encountering an `InvalidParameterError`, the middleware or application should just return an empty 400 response to indicate that the request will not be processed, rather than allowing the exception to throw a 500 error. This happens in [Rails][], so when a particular request/IP combo is being thwarted by `Rack::Attack`, you'll still get an error that ends up being a bunch of noise in your error monitoring or logging system. To address this, `rescue` the errors that could cause this particular situation and respond with a clean 400 error.

[Rack::QueryParser#parse_nested_query]: https://www.rubydoc.info/github/rack/rack/Rack%2FQueryParser:parse_nested_query
[Rails]: https://github.com/rails/rails/blob/fdd5894aa985651fb99e0ef1c2d3f3f604b2f659/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb#L23

### DIFF
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -22,6 +22,15 @@ module Rack
     class MissingStoreError < Error; end
     class IncompatibleStoreError < Error; end
 
+    INVALID_PARAMETER_ERRORS = if defined?(Rack::RELEASE)
+                                 [
+                                   Rack::Utils::InvalidParameterError,
+                                   Rack::QueryParser::InvalidParameterError
+                                 ]
+                               else
+                                 []
+                               end
+
     autoload :Check,                'rack/attack/check'
     autoload :Throttle,             'rack/attack/throttle'
     autoload :Safelist,             'rack/attack/safelist'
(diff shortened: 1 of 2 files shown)


grzuy:
Hi @tubbo , welcome!

Can you please give a bit more detail on what's is the problem you're trying to solve in your particular case. What you were experiencing that you're trying to solve?

Trying to understand how this fits into a rack-attack responsibility.

Thanks.

tubbo:
Hey @grzuy, the issue here is that when invalid parameter errors are encountered, rack-attack will just allow the exception to raise, and that eventually boils down to a 500 error. This exception is supposed to be caught by a Rack app so the app can return a 400 response, but since Rack::Attack allows the error to be thrown, that can never happen. An active attack on a website results in a torrent of these errors spamming Sentry, logs, and/or any other monitoring tools that capture exceptions from the app. Mostly just annoying and unnecessary.

tubbo:
@grzuy you cool with this or do you think it's not a rack-attack responsibility? I'm cool with closing if you are.

vipulnsward:
Experienced these recently. During attacks the 500s spike up a lot + I think this leads to also not throttling the requests as well?

PR #502, opened 2020-10-21
Lock rubocop-ast version.

simi:
- newer version seems to not be compatible with used rubocop version

This should fix CI for now, later we can upgrade whole rubocop stack as well.


ioquatix:
Thanks for this.

grzuy:
I ended up taking the opportunity to give rubocop an version jump to earlier version that doesn't have the AST crash: https://github.com/rack/rack-attack/commit/cbae022df107b58997ac11f15a442a1631811b39.

Thanks @simi anyway for this!

Issue #508, opened 2020-12-15
Ruby 2.7 deprecation warning with ActiveSupport(v6.1.0) RedisCacheStore


fukayatsu:
## Steps to Reproduce

1. Start redis
    - e.g. `docker run -p 6379:6379 redis`
2. Run below script on ruby 2.7.2
    ```ruby
    # rack_attack_debug.rb
    
    require 'bundler/inline'
    
    gemfile do
      source 'https://rubygems.org'
      gem 'activesupport', '6.1.0'
      gem 'rack-attack', '6.3.1'
      gem 'redis', '4.2.5'
    end
    
    Warning[:deprecated] = true
    
    require 'active_support'
    require 'active_support/cache/redis_cache_store'
    Rack::Attack.cache.store = ActiveSupport::Cache::RedisCacheStore.new
    
    key = 'rack_attack_debug'
    Rack::Attack.cache.store.delete(key)
    
    # https://github.com/rack/rack-attack/blob/v6.3.1/lib/rack/attack/store_proxy/redis_cache_store_proxy.rb#L13
    Rack::Attack.cache.store.increment(key, 1, expires_in: 30) #=> RedisCacheStore#write with expires_in
    Rack::Attack.cache.store.increment(key, 1, expires_in: 30) #=> RedisCacheStore#increment
    puts "#{key}: #{Rack::Attack.cache.store.read(key)}"
    ```

## Expected

```
$ ruby -v
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin19]

$ ruby rack_attack_debug.rb
rack_attack_debug: 2
```

## Actual

```
$ ruby -v
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin19]

$ ruby rack_attack_debug.rb
/Users/fukayatsu/.rbenv/versions/2.7.2/lib/ruby/2.7.0/delegate.rb:83: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/fukayatsu/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/activesupport-6.1.0/lib/active_support/cache/strategy/local_cache.rb:116: warning: The called method `increment' is defined here
rack_attack_debug: 2
```

PR #509, opened 2020-12-15
Fix Ruby 2.7 kwargs warning in RedisCacheStoreProxy

fukayatsu:
fix https://github.com/rack/rack-attack/issues/508


grzuy:
Thank you!

PR #516, opened 2021-01-28
Add Truffleruby head to CI

gogainda:


gogainda:
there were some problems with downloading ubuntu, will close for now

PR #520, opened 2021-02-18
Update readme documentation

pinkahd:
The `blocklisted_callback` and  `throttled_callback` are no longer available
and have been removed. The correct way to specify those callbacks are via
`blocklisted_response` and `throttled_response`.

Also those callbacks are now getting the `env` send as the argument instead
of the `request` object.

PR #540, opened 2021-04-06
corrects blocklisted_callback and throttled_callback (should be _resp…

jasonfb:
…onse)

I think this is right


jasonfb:
actually I 'spoke' too soon; I see this is already pull-requested in https://github.com/rack/rack-attack/pull/520

looks like @pinkahd already got this

PR #548, opened 2021-06-07
Notify discriminator for tracked requests

dsalahutdinov:
Hi!

I am working on the monitoring solution for rack attack [here](https://github.com/dsalahutdinov/yabeda-rack-attack). It listens to events and [increments specific counters](https://github.com/dsalahutdinov/yabeda-rack-attack/blob/master/lib/yabeda/rack/attack/notifier.rb#L31-L33) for any rule type (throttle, blacklist, whitelist, and track).

For now, if I write the following track rule - I will have any requests matching. But there would not be discriminator passed to the event. I think tracking requests in context is an in-demand feature. It will be great to pass additional information about the block-result value.

```
Rack::Attack.track('requests by user agent family') do |req|
  UserAgentParser.parse(req.user_agent).family
end
```

Here is how nice we can visualize the data, collected in context of result:

Screenshot 2021-06-07 at 23 49 05

I think this feature opens up new opportunities to do initial research before setting up any rule for production. We can first track queries for some time to collect some statistics for proposal rule this way. Great example:

```
Rack::Attack.track('tracks requests per user') do |req|
  req.env['warden']&.user&.id
end
```
Then collect any statistics to analyze how many requests every user does.

Then implement throttling with something like this:
```
Rack::Attack.throttle('requests per user', limit: 10, period: 1.minute) do |req|
  req.env['warden']&.user&.id
end
```

My proposal pull-request keeps the "tracks" feature logic the same and is fully compatible for all the listeners.

### DIFF
--- a/lib/rack/attack/check.rb
+++ b/lib/rack/attack/check.rb
@@ -15,6 +15,7 @@ module Rack
         block.call(request).tap do |match|
           if match
             request.env["rack.attack.matched"] = name
+            request.env['rack.attack.match_discriminator'] = match
             request.env["rack.attack.match_type"] = type
             Rack::Attack.instrument(request)
           end


lib/rack/attack/check.rb
```diff
@@ -15,6 +15,7 @@ def matched_by?(request)
         block.call(request).tap do |match|
           if match
             request.env["rack.attack.matched"] = name
+            request.env['rack.attack.match_discriminator'] = match
```
grzuy, line 18: So this might be a bit odd because `Check`'s `block` are mostly intended to return `true` or `false`, not a "discrimiator" as in the case of throttling, right?

I see how the `#track` method makes this a bit confusing... given the fact that behind the scenes it used the `Throttle` or `Check` based on keyword arguments presence...

grzuy (commented):
Hi @dsalahutdinov,

Thank you for the pull request and the detailed description!

lib/rack/attack/check.rb
dsalahutdinov: Hi, @grzuy ! Thank you for your feedback.

Sure, you are right!. It's kind of confusing for `Check` rule to return a non-boolean value.

Motivation for proposed changes is a possibility of initial research. Before rolling out any throttling rule to production, it's would great to estimate how this rule performs (without affecting production).

If I want to test the following rule (before turning it on):
```ruby
  Rack::Attack.throttle('req/ip', :limit => 10, :period => 1.hour) do |req|
    req.ip
  end
```
I could track it, collect any real values and decide how strictly my throttling rules should be:

```
  Rack::Attack.track('requests by ip address') do |req|
    req.ip
  end
```
This way lets me make data-driven decisions. For that purpose `Track` feature fits perfectly, I think.

What do you think, It should look like? What do you think if we refactor `Check` class a bit, rename it to `Track` to make gem internals less confusing, and give my proposal a go?

grzuy: In that case wouldn't you want to

```rb
  Rack::Attack.track('req/ip', limit: 10, period: 1.hour) do |req|
    req.ip
  end
```

so that it really uses the `Throttle` (instead of `Check`) behind the scenes, in that case getting the `match_discriminator` in the request env?

dsalahutdinov: Yep, it is. My point is to initially use a non-affecting track with the same rule. It helps to determine actual traffic depending on a discriminator without affecting production.
Then, use the valid data-drive throttle limit and period for turning on to a 'trottle'.

ioquatix:

grzuy:
@dsalahutdinov Thanks again for this PR.

As is, and considering my previous comment unfortunately I think not something we want to merge. Trying to avoid more confusion to an already bit confusing "track" feature.

I do understand and I think it's interesting what you're trying to solve. I think it might be something like define rules in kind of a "dry-run" mode if I understand correctly.

Not sure `Rack::Attack.track`, as per the actual implementation is the best for that.

Please feel free to open a new issue asking for a "Dry-run mode for rules" and see whether a plan to implement that consistently with how the gem works as of now comes up from discussion.

Thanks.

PR #554, opened 2021-09-06
Add Truffleruby head to CI

gogainda:


.github/workflows/build.yml
```diff
@@ -19,19 +19,14 @@ jobs:
         ruby:
           - 3.1.2
           - 3.0.4
-          - 2.7.6
-          - 2.6.10
-          - 2.5.8
```
santib, line 24: I don't think this PR should remove other ruby/rails versions from the CI

.github/workflows/build.yml
```diff
@@ -19,19 +19,14 @@ jobs:
...
+          - truffleruby-head
```
santib, line 22: I don't think we want to test against head, we don't even do that with MRI. Maybe we could add some stable versions instead

grzuy: Agree.

grzuy:
Sorry this became a bit stale because of my delay.

Feel free to re-open with a stable version of truffleruby.

Thanks @gogainda.

PR #556, opened 2021-10-20
Add support for Rails 7.0

zarqman:
This PR adds support for ~~both Dalli 3.0 and~~ Rails 7.0.

Per rails/rails#43492, in Rails 7 the API for MemCacheStore `read()` has an intentional backwards incompatible change in that `raw: true` is now required on `read()` or nil is returned.

Rails 7 is still in alpha, so this could be considered premature, but we're already working to validate a couple of apps on Rails 7 and found this. Thought others might be doing the same and hoping to save others the same work.

Since the Rails 7 change is dependent on Dalli 3 in the appraisal, it seemed best to combine these, but will split PRs if needed.

I tried to update the github build.yml as well. Apologies if that's an overreach--happy to remove if it's too much.


grzuy:
Hi @zarqman, thank you very much for this!

If I understand correctly these are two independent changes right? If so, can we please have separate pull requests for each commit?

Thanks again!!

grzuy:
@zarqman Thanks a lot for this!

> Since the Rails 7 change is dependent on Dalli 3 in the appraisal, it seemed best to combine these, but will split PRs if needed.

Can we have separate PR for testing against Dalli 3 first please?
Then we have this PR be adding support for Rails 7 only?

Thanks again!

zarqman:
> Can we have separate PR for testing against Dalli 3 first please? Then we have this PR be adding support for Rails 7 only?

Dalli 3 portion is now in PR #557.

zarqman:
This has been rebased and now just contains the Rails 7 updates.

Appraisals
```diff
@@ -17,6 +17,11 @@ appraise "rack_1" do
   gem "rack-test", ">= 0.6"
 end
 
+appraise 'rails_7-0' do
+  gem 'railties', '~> 7.0.0.a'
+  gem 'dalli', '~> 3.0'
```
grzuy, line 22: Why is that we're adding `dalli` explicitly on this `rails 7` appraise?

Not consistent to how we are testing the rest of the rails versions.

grzuy (commented):
Thanks again!

Appraisals
zarqman: Because otherwise dalli isn't included and the memcache+dalli tests are skipped on Rails 7, but the specific behavior change we're testing and correcting for is in Rails 7's use of memcache+dalli.

grzuy: Gotcha.

The thing is we're not using `rails` appraisals for testing integration with the rest of the gems.

That's what `dalli3` appraisal will eventually be for, when rails 7 is out of the beta. If you check `gemfiles/dalli3.gemfile.lock` right now is testing with `activesupport` 6.1, but will switch to `7.0` when final release.

In the meantime I prefer we add an additional separate appraisal just for the purpose of testing against rails beta and call it out with a comment that could be removed once rails 7 is final.

grzuy: Updated myself.

PR #558, opened 2021-11-19
docs: update link to rack spec in README

agbaber:
The link for the Rack app interface was broken.

Updated the link so it now resolves to the intended documentation.


grzuy:
Nice! :eyes: 

Thank you!!

Issue #561, opened 2022-01-12
Is Rack::Attack.reset! supposed to clear .blocklist and .safelist?


johlym:
I've been banging my head on this one all afternoon trying to figure out why my tests for Rack::Attack behavior aren't reproducing what I'm expecting.

One of my `describe`s runs 
```rb
before(:each) do
  Rack::Attack.safelist_ip(known_address_block)
  Rack::Attack.blocklist_ip(catch_all_block)
end
```

which is fine. The tests thereafter do exactly what they're supposed to do.

My other `describe` has no such `before(:each)`, expects no `.blocklist` or `.safelist` to be defined, and if it gets to go first in random order, it works great. If it ends up second, the `.blocklist`/`.safelist` generation action is carried over from the first, even though I have

```rb
after(:each) do
  Rack::Attack.reset!
end
```

at the root of my test suite, being called after every test.

What am I missing? Does `Rack::Attack.reset!` not clear the block-/safelists? Is there a way I can? My workaround is to manually define `Rack::Attack.safelist_ip(catch_all_block)` to take advantage of precedence, but that's not how Rack::Attack is initialized in the real world.

johlym:
As it turns out, `.clear_configuration` is the tool for the job, here. It's not documented, so far as I can tell, so I might submit a PR for the README to include it.

johlym:
Addressed in #562.

PR #562, opened 2022-01-17
Update README to mention .clear_configuration

johlym:
Adds a line to the `Test case isolation` section of `README.md` about `.clear_configuration`. This manifested after I ran into the issue described in #561 with blocklist and safelist states not being cleared between test cases using `.reset!`.


grzuy:
Thank you @johlym!! :clap:

PR #565, opened 2022-01-28
Include LICENSE in gem build

orhantoy:
Hi 👋 

Including the LICENSE in gem builds is helpful when generating a report including licenses from all your dependencies.
`rack/rack` [does it](https://github.com/rack/rack/blob/c1e5fbbb59101c039e8b657c8052e152c572d5ac/rack.gemspec#L20-L21) for instance.


grzuy:
Nice, thanks!

PR #573, opened 2022-02-19
Lower case headers.

ioquatix:
See https://github.com/rack/rack/pull/1812 for more details.

### DIFF
--- a/lib/rack/attack/configuration.rb
+++ b/lib/rack/attack/configuration.rb
@@ -5,7 +5,7 @@ require "ipaddr"
 module Rack
   class Attack
     class Configuration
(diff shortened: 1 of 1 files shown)


jeremyevans:
Seems fine to me.  May be more backwards compatible to only use the lowercase headers for Rack > 2.3.0, but I'm not sure it's worth doing that.

ioquatix:
@grzuy we want to merge this to test @jeremyevans 's upstream PR, but if this isn't acceptable finally (i.e. we decide to back out of @jeremyevans's PR) I'll revert this.

grzuy (commented):
Looks good :heavy_check_mark: 

I don't interpret this change being backwards-incompatible. These headers should already be treated as case-insesitive by any HTTP client parsing/reading them.

Thanks for working on this in rack @ioquatix and @jeremyevans.

We could also make some updates in the `README` I guess.

PR #574, opened 2022-03-01
Document an issue that can cause flaky tests

tuzz:


grzuy:
Hi @tuzz,

Thanks for the taking the time to describe the issue you faced and opening the PR.

I think it best fits a "General" or "Show and tell" post in the Discussions tab. Please feel free to post there. There's the benefit of the interaction and comments from others there and also future comments updating the solution/workaround if things changes because of changes in the gem in the future. Which is more difficult to track and update if the workaround is in the README i think.

Thanks again!

PR #575, opened 2022-03-08
fix(documentation): Updating README with non-deprecated configuration.

adriancb:
Hi Team,

First of all, thank you for your continued effort in maintaining this gem. It's appreciated.

This PR attempts to fix the documentation given recent config deprecation.


grzuy:
Thank you very much!!

PR #579, opened 2022-04-28
Replace git.io URL

kyoshidajp:
Hi,

git.io URL has been deprecated, and it will stop redirecting on April 29, 2022.

For more detail:
https://github.blog/changelog/2022-04-25-git-io-deprecation/


grzuy:
Thank you very much!

PR #586, opened 2022-08-03
Support rack 3

### LINKED ISSUE #596: Rack 3 Support
Hi all,

I was looking to upgrade to Rack 3.0, but can't as rack-attack's `6.6.1` depends on `(rack>= 1.0, <3)`. I know there were some major changes to the Rack 3.0 upgrade, so I figured I would make an issue to get Rack 3 compatibility on the radar.

ioquatix:
We should probably also update CI, but we've been testing `rack-attack` from the `rack` repo (external tests) and all the tests are passing. So this should probably be good.

--- 

Closes #596


gemfiles/rack_3.gemfile
```diff
@@ -0,0 +1,8 @@
+# This file was generated by Appraisal
+
+source "https://rubygems.org"
+
+gem "rack", "~> 3.0", git: "https://github.com/rack/rack.git"
```
jeremyevans, line 5: Not sure you need the `~> 3.0`, but maybe you just delete the `git` option after the Rack 3 official release?

gemfiles/rack_3.gemfile
```diff
@@ -0,0 +1,8 @@
...
+gem "rack-test", git: "https://github.com/rack/rack-test.git"
```
jeremyevans, line 6: Unless you have a good reason not to, I would drop the `git` option for rack-test.

jeremyevans (changes requested):

gemfiles/rack_3.gemfile
ioquatix: Yes, that's correct.

gemfiles/rack_3.gemfile
ioquatix: It doesn't seem to work without it.

gemfiles/rack_3.gemfile
```diff
@@ -0,0 +1,8 @@
+# This file was generated by Appraisal
```
grzuy, line 1: Normally this file is autogenerated by adding a new entry to `Appraisals` file and running `appraisal install`.

gemfiles/rack_3.gemfile
grzuy: It seems this is needed here because we're out of date in the gemspec with regards to `rack-test`.

This PR should fix that https://github.com/rack/rack-attack/pull/587, so that we don't need this here.

grzuy (commented):
Cool, thank you!

I think to have the test pass we would additionally need to point to versions of actionpack that actually compatible with rack 3.

gemfiles/rack_3.gemfile
grzuy: > but maybe you just delete the `git` option after the Rack 3 official release?

Done.

gemfiles/rack_3.gemfile
grzuy: Done in https://github.com/rack/rack-attack/pull/586/commits/5d3da4975a8c8d424ef84acdf672684ff9ba04ab.

grzuy:
> I think to have the test pass we would additionally need to point the rack_3 appraisal to a version of actionpack that actually compatible with rack 3.

This depends on https://github.com/rails/rails/pull/45741 being merged first probably.

.github/workflows/build.yml
```diff
@@ -23,6 +23,7 @@ jobs:
           - 2.6.10
           - 2.5.8
         gemfile:
+          - rack_3
```
grzuy, line 26: We most probably would need to add a few exclusions of rack 3 in combination with some ruby versions down below in this file later.

halostatue:
What's blocking this from being merged?

jeremyevans:
I think CI needs to be green for the PR before we can consider merging.  No issues with the PR itself, though.

ioquatix:
It looks like this is blocked on https://github.com/rails/rails/pull/46594 and indirectly https://github.com/rails/rails/pull/46206 - @rafaelfranca FYI.

ioquatix:
This gem is too heavily dependent on Rails (`Rails.cache` and `Rails::Railtie`) to work independently of it. I could not get tests to pass with only ActiveSupport loaded.

smridge:
From what I'm seeing, this will work if rails `edge` is used now that the version restriction has been relaxed [here](https://github.com/rails/rails/commit/859b526c5b9f1e516df3fc1b388c949da3fa9c4e)

fayeroske:
Any updates here? Similarly blocked trying to remediate CVE-2022-30122 by upgrading to rack 3.0.

xxswingxx:
Since https://github.com/rails/rails/pull/46206 was merged, it's just https://github.com/rails/rails/pull/46594 blocking this one?

ioquatix:
Sorry, I totally forgot about this PR. I'll take a look at it in the next few days.

nedim-symphony:
@ioquatix  Hello, any updates on this?

tsmartt:
Same, any update? And thanks for the library!

tenderlove (approved):

PR #604, opened 2022-12-16
Add note about cache stores and in-memory caches.

nateberkopec:
This was part of the "lessons learned" from Bannerbear's DDoS attack: https://www.bannerbear.com/blog/how-to-survive-a-ddos-attack/

I think perhaps many people are blindly copying this line from the readme:

```ruby 
Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new # defaults to Rails.cache
```

There really isn't a reason to have this line in the Readme, because the default is OK. We should include an example of what most apps will probably want to do, which is to set this to a Redis or Memcached store, not a MemoryStore, which is probably appropriate only for local-only toy apps.


grzuy (approved):
README changes make sense to me.

Thank you @nateberkopec!

PR #627, opened 2023-10-14
ci: no longer test against rack 1

grzuy:


santib (approved):
💯

ioquatix:
I think totally reasonable.

PR #630, opened 2023-10-16
Fix #588 don't fail if request.ip is missing

### LINKED ISSUE #588: Use of safelist_ip causes error when IP address is not set
I have a rake task that simulates a request to my server. I use this as part of a deploy script as a sanity check to ensure the code I'm deploying is functional.

```ruby
# For deployment scripts, to check that the homepage is functional.
namespace :server_status do
  task check_homepage: :environment do
    host = 'example.com'
    rack_env = {
      'rack.input'     => StringIO.new,
      'PATH_INFO'      => '/',
      'REQUEST_METHOD' => 'GET',
      'HTTP_HOST'      => host,
      'HTTPS'          => 'on',
    }
    status, _headers, _body = Rails.application.call(rack_env)
    if status != 200
      raise "Request to #{host} failed with status #{status}!"
    else
      puts "Request to #{host} succeeded."
    end
  end
end
```

When I add `safelist_ip` call to my rack-attack config, this starts to fail:

```
IPAddr::AddressFamilyError: address family must be specified
  from ipaddr.rb:565:in `initialize'
  from rack-attack (6.6.1) lib/rack/attack/configuration.rb:68:in `new'
  from rack-attack (6.6.1) lib/rack/attack/configuration.rb:68:in `block in safelist_ip'
  from rack-attack (6.6.1) lib/rack/attack/check.rb:15:in `matched_by?'
  from rack-attack (6.6.1) lib/rack/attack/configuration.rb:80:in `block in safelisted?'
  from rack-attack (6.6.1) lib/rack/attack/configuration.rb:80:in `any?'
  from rack-attack (6.6.1) lib/rack/attack/configuration.rb:80:in `safelisted?'
  from rack-attack (6.6.1) lib/rack/attack.rb:109:in `call'
```

Adding `REMOTE_ADDR` to `rack_env` above makes it succeed again, but it would be nice if this wouldn't blow up rack-attack.

santib:
From https://github.com/rack/rack-attack/issues/588 we can see that if for some reason the `request.ip` is missing, `rack-attack` will try to do `IPAddr.new(nil)` which raises `*** IPAddr::AddressFamilyError Exception: address family must be specified`.

With this PR we only call IPAddr.new if we have a not empty string

### DIFF
--- a/spec/acceptance/safelisting_ip_spec.rb
+++ b/spec/acceptance/safelisting_ip_spec.rb
@@ -12,6 +12,12 @@ describe "Safelist an IP" do
   end
 
   it "forbids request if blocklist condition is true and safelist is false" do
+    get "/admin", {}, "REMOTE_ADDR" => "1.2.3.4"
+
+    assert_equal 403, last_response.status
+  end
+
+  it "forbids request if blocklist condition is true and safelist is false (missing IP)" do
     get "/admin", {}, "REMOTE_ADDR" => ""
 
     assert_equal 403, last_response.status
--- a/spec/acceptance/safelisting_ip_spec.rb
+++ b/spec/acceptance/safelisting_ip_spec.rb
@@ -12,7 +12,7 @@ describe "Safelist an IP" do
   end
 
   it "forbids request if blocklist condition is true and safelist is false" do
-    get "/admin", {}, "REMOTE_ADDR" => "1.2.3.4"
+    get "/admin", {}, "REMOTE_ADDR" => ""
 
     assert_equal 403, last_response.status
   end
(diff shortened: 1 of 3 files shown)


spec/acceptance/safelisting_ip_spec.rb
```diff
@@ -12,7 +12,7 @@
   end
 
   it "forbids request if blocklist condition is true and safelist is false" do
-    get "/admin", {}, "REMOTE_ADDR" => "1.2.3.4"
+    get "/admin", {}, "REMOTE_ADDR" => ""
```
grzuy, line 15: Same here?

grzuy (commented):
Thank you!

grzuy (approved):
:clap:

PR #632, opened 2023-10-24
ci: test against ActiveSupport::RedisCacheStore v5

santib:
Add gemfile to prevent regressions in ActiveSupport 5 which is actively supported by `rack-attack`

Note:
- Changes in other gemfiles were done by Appraisals


grzuy (approved):
Thank you!

Just realized the `active_support_redis_cache_store` appraisal points to activesupport v6. We're not testing against v7.

We may benefit from having one for each.

PR #634, opened 2023-10-25
ci: automatically run with latest ruby patch version

santib:
With this change we don't need to care about patch versions

Note: I'm using strings as suggested in https://github.com/ruby/setup-ruby#supported-version-syntax. If you don't use the commas, most versions just work but `3.0` fails for some strange reason

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,12 +17,12 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - 3.2.2
-          - 3.1.4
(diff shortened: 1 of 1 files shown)


grzuy (approved):
Nice quality of life improvement!
Didn't know setup-ruby had this feature.

Thank you!

grzuy:
>  commas,

You meant quotes I assume.

santib:
> > commas,
> 
> You meant quotes I assume.

Yes, exactly 😅

PR #639, opened 2023-11-16
System Hardening Phase 1: Catch errors at top level

johnnyshields:
In PR https://github.com/rack/rack-attack/pull/577, @grzuy asked me to break it into a series of smaller PRs.

My plan is as follows:
- [x] **This PR**:
   - [x] Move error catching to top-level of `Rack::Attack`. This is mandatory to do, because we need to have Rack::Attack be aware of errors in order to implement advanced error handling/fault-tolerance mechanisms in subsequent PRs.
   - [x] Remove the `rescuing` blocks from `Dalli/RedisProxy` classes.
   - [x] Add "Rack::Attack.allowed_errors" config which defaults to `Dalli::DalliError` and `Redis::BaseError`. This config is useful if you are implementing a custom store class.

Future PRs (will raise the next one after this one is merged)

- [ ] PR 2:
  - [ ] Add Rack::Attack.error_handler which takes a Proc for custom error handling.
  - [ ] Refactor Rack::Attack a bit so the error_handler can select whether to throttle or block.
- [ ] PR 3: Add Rack::Attack.calling? so Rails Cache error hander can raise its error for Rack::Attack.
- [ ] PR 4: Add Rack::Attack.failure_cooldown config. This temporarily disables Rack::Attack after an error occurs (including allowed errors), to prevent cache connection latency.

---

Note: This PR will affect the small number of users are directly using the `Dalli/RedisProxy` classes (or subclassing them, etc.) since the methods of these classes now raise `Dalli::DalliError` /`Redis::BaseError` rather than returning `nil`. It should probably be released as a major version for this reason.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ See the [Backing & Hacking blog post](https://www.kickstarter.com/backing-and-ha
 - [Customizing responses](#customizing-responses)
   - [RateLimit headers for well-behaved clients](#ratelimit-headers-for-well-behaved-clients)
 - [Logging & Instrumentation](#logging--instrumentation)
+- [Custom Error Handling](#custom-error-handling)
 - [Testing](#testing)
 - [How it works](#how-it-works)
   - [About Tracks](#about-tracks)
@@ -400,11 +401,21 @@ ActiveSupport::Notifications.subscribe(/rack_attack/) do |name, start, finish, r
 end
 ```
 
+## Custom Error Handling
+
+You may specify the list of internal errors to allow (i.e. not raise an error)
+as either an array of Class and/or String values.
+
+```ruby
+# in initializers/rack_attack.rb
+Rack::Attack.allowed_errors += [MyErrorClass, 'MyOtherErrorClass']
+```
+
 ## Testing
 
-A note on developing and testing apps using Rack::Attack - if you are using throttling in particular, you will
-need to enable the cache in your development environment. See [Caching with Rails](http://guides.rubyonrails.org/caching_with_rails.html)
-for more on how to do this.
+When developing and testing apps using Rack::Attack, if you are using throttling in particular,
+you must enable the cache in your development environment. See
+[Caching with Rails](http://guides.rubyonrails.org/caching_with_rails.html) for how to do this.
 
 ### Disabling
 
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -63,10 +63,10 @@ module Rack
       end
 
       def allow_error?(error)
-        allowed_errors&.any? do |ignored_error|
-          case ignored_error
-          when String then error.class.ancestors.any? {|a| a.name == ignored_error }
-          else error.is_a?(ignored_error)
+        allowed_errors&.any? do |allowed_error|
+          case allowed_error
+          when String then error.class.ancestors.any? {|a| a.name == allowed_error }
+          else error.is_a?(allowed_error)
           end
         end
       end
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -32,8 +32,14 @@ module Rack
     autoload :Fail2Ban,             'rack/attack/fail2ban'
     autoload :Allow2Ban,            'rack/attack/allow2ban'
 
+    DEFAULT_ALLOWED_ERRORS = %w[Dalli::DalliError Redis::BaseError].freeze
+
     class << self
-      attr_accessor :enabled, :notifier, :throttle_discriminator_normalizer
+      attr_accessor :enabled,
+                    :notifier,
+                    :throttle_discriminator_normalizer,
+                    :allowed_errors
+
       attr_reader :configuration
 
       def instrument(request)
@@ -57,6 +63,15 @@ module Rack
 
       def reset!
         cache.reset!
+      end
+
+      def allow_error?(error)
+        allowed_errors&.any? do |ignored_error|
+          case ignored_error
+          when String then error.class.ancestors.any? {|a| a.name == ignored_error }
+          else error.is_a?(ignored_error)
+          end
+        end
       end
 
       extend Forwardable
@@ -86,7 +101,10 @@ module Rack
       )
     end
 
-    # Set defaults
+    # Set class defaults
+    self.allowed_errors = DEFAULT_ALLOWED_ERRORS.dup
+
+    # Set instance defaults
     @enabled = true
     @notifier = ActiveSupport::Notifications if defined?(ActiveSupport::Notifications)
     @throttle_discriminator_normalizer = lambda do |discriminator|
@@ -128,6 +146,8 @@ module Rack
         configuration.tracked?(request)
         @app.call(env)
       end
+    rescue StandardError => error
+      self.class.allow_error?(error) ? @app.call(request.env) : raise(error)
     end
   end
 end
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -32,13 +32,10 @@ module Rack
     autoload :Fail2Ban,             'rack/attack/fail2ban'
     autoload :Allow2Ban,            'rack/attack/allow2ban'
 
-    DEFAULT_ALLOWED_ERRORS = %w[Dalli::DalliError Redis::BaseError].freeze
-
     class << self
       attr_accessor :enabled,
                     :notifier,
-                    :throttle_discriminator_normalizer,
-                    :allowed_errors
+                    :throttle_discriminator_normalizer
 
       attr_reader :configuration
 
@@ -97,12 +94,11 @@ module Rack
         :safelists,
         :blocklists,
         :throttles,
-        :tracks
+        :tracks,
+        :allowed_errors,
+        :allowed_errors=
       )
     end
-
-    # Set class defaults
-    self.allowed_errors = DEFAULT_ALLOWED_ERRORS.dup
 
     # Set instance defaults
     @enabled = true
--- a/rack-attack.gemspec
+++ b/rack-attack.gemspec
@@ -32,6 +32,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "bundler", ">= 1.17", "< 3.0"
   s.add_development_dependency 'minitest', "~> 5.11"
   s.add_development_dependency "minitest-stub-const", "~> 0.6"
+  s.add_development_dependency 'rspec-mocks', '~> 3.11.0'
   s.add_development_dependency 'rack-test', "~> 2.0"
   s.add_development_dependency 'rake', "~> 13.0"
   s.add_development_dependency "rubocop", "1.12.1"
--- a/spec/acceptance/error_handling_spec.rb
+++ b/spec/acceptance/error_handling_spec.rb
@@ -1,0 +1,93 @@
+# frozen_string_literal: true
+
+require_relative "../spec_helper"
+
+describe "error handling" do
+
+  let(:store) do
+    ActiveSupport::Cache::MemoryStore.new
+  end
+
+  before do
+    Rack::Attack.cache.store = store
+
+    Rack::Attack.blocklist("fail2ban pentesters") do |request|
+      Rack::Attack::Fail2Ban.filter(request.ip, maxretry: 0, bantime: 600, findtime: 30) { true }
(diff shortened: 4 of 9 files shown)


johnnyshields:
@grzuy ready for your review.

README.md
```diff
@@ -400,11 +401,21 @@ ActiveSupport::Notifications.subscribe(/rack_attack/) do |name, start, finish, r
...
 ```
 
+## Custom Error Handling
+
+You may specify the list of internal errors to allow (i.e. not raise an error)
+as either an array of Class and/or String values.
+
+```ruby
+# in initializers/rack_attack.rb
+Rack::Attack.allowed_errors += [MyErrorClass, 'MyOtherErrorClass']
```
grzuy, line 411: Not strong, thoughts on naming `ignored_errors`  instead `allowed_errors`? :thinking:

rack-attack.gemspec
```diff
@@ -32,6 +32,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "bundler", ">= 1.17", "< 3.0"
   s.add_development_dependency 'minitest', "~> 5.11"
   s.add_development_dependency "minitest-stub-const", "~> 0.6"
+  s.add_development_dependency 'rspec-mocks', '~> 3.11.0'
```
grzuy, line 35: Haven't looked at the in more detail but would be great if we wouldn't need to add extra dep.

spec/acceptance/error_handling_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    it 'can ignore error ancestor as Class' do
+      Rack::Attack.allowed_errors = [StandardError]
+
+      get "/", {}, "REMOTE_ADDR" => "1.2.3.4"
+
+      assert_equal 200, last_response.status
+    end
+
+    it 'can ignore error as String' do
+      Rack::Attack.allowed_errors = %w[RuntimeError]
```
grzuy, line 50: Can we just go with `["RuntimeError"]` so it's more obvious what type it is?

grzuy (commented):
@johnnyshields Thanks for breaking it up and opening smaller PR.

Left some comments after first pass.

Lets please update PR title to something that explains what this does for end users. Would that be "Allows users to specify list of ignored errors" or something similar?

grzuy:
@santib Would be good to have your thoughts on this one too :pray: :-)

README.md
johnnyshields: @grzuy the reason for "allowed" here is that when these errors occur, RackAttack "allows" the request. This naming will be made more obvious in the next PR, which allows you to add error handlers which throttle or block based on certain errors.

rack-attack.gemspec
johnnyshields: We need it. In the tests you can see I am stubbing certain methods to raise errors. There is no other way to do this in the specs b/c these error would only occur if certain infra such as Redis is down. The only alternative would be to write monkey patches in the tests, which is a far worse/messy option that using RSpec's widely-used mock library (which has minitest integration)

spec/acceptance/error_handling_spec.rb
johnnyshields: The Rubocop style guide here says a one-element array can be either `['RuntimeError']` or `%w[RuntimeError]`, but a two-element array must use `%w[]` e.g. `%w[RuntimeError OtherError]`

Since I'm using both 1 and 2 element arrays this test (as well as elsewhere), I think it is better to use `%w[ ]` consistently everywhere. Please confirm if you'd like me to (a) keep as-is, (b) change only one-element arrays, (c) change both 1 and 2-element arrays.

See: https://docs.rubocop.org/rubocop/cops_style.html#stylewordarray

lib/rack/attack.rb
```diff
@@ -32,8 +32,14 @@ class IncompatibleStoreError < Error; end
...
     autoload :Allow2Ban,            'rack/attack/allow2ban'
 
+    DEFAULT_ALLOWED_ERRORS = %w[Dalli::DalliError Redis::BaseError].freeze
+
     class << self
-      attr_accessor :enabled, :notifier, :throttle_discriminator_normalizer
+      attr_accessor :enabled,
+                    :notifier,
+                    :throttle_discriminator_normalizer,
+                    :allowed_errors
```
santib, line 41: I see this as a new configuration, so I'd move it to `lib/rack/attack/configuration.rb`. See `throttled_responder` as an example. WDYT? Do you see any reason why to have it here instead?

lib/rack/attack.rb
```diff
@@ -128,6 +146,8 @@ def call(env)
         configuration.tracked?(request)
         @app.call(env)
       end
+    rescue StandardError => error
+      self.class.allow_error?(error) ? @app.call(request.env) : raise(error)
```
santib, line 150: So, here we are rescuing `StandardError` so we can allow strings in the `allowed_errors` configuration, right? Is there any reason why to not do this?

```ruby
rescue allowed_errors => error
  @app.call(request.env)
end
```

because I prefer the simplicity of this alternative if possible

lib/rack/attack.rb
johnnyshields: @santib done.

lib/rack/attack.rb
johnnyshields: I did it this way so we can define the `allowed_errors` as strings, and therefore not have a dependency on `Dalli` or `Redis` constants being loaded. Note the `allow_error?` method is as follows:

```ruby
      def allow_error?(error)
        allowed_errors&.any? do |ignored_error|
          case ignored_error
          when String then error.class.ancestors.any? {|a| a.name == ignored_error }
          else error.is_a?(ignored_error)
          end
        end
      end
```

rack-attack.gemspec
santib: I get your point, but maybe we can change [where this gem is used](https://github.com/rack/rack-attack/pull/639/files#diff-4da0c4f87f4ac598efc2c860433f206defbc573cf17f7f5f37de77faf3a54f73R21) for something similar to how [exceptions are currently being forced in tests](https://github.com/rack/rack-attack/blob/main/spec/integration/offline_spec.rb#L31), and we avoid introducing a new dependency? Or are there other places where `rspec-mocks` is needed?

lib/rack/attack.rb
santib: >not have a dependency on Dalli or Redis constants being loaded

Agree on this.

Not really sold on the `allow_error?` method, I'd prefer to use Ruby's exception handling instead, while finding other way to solve the need to not load `Dalli`/`Redis` constants.

I know you didn't like it before, but one way could be the `Rack::Attack::CacheError` class, but there could be others, for example having this class wrap and re-raise the original error, maybe? Or maybe having the defaults defined by each proxy class which also checks for `defined?(Dalli)` [as they already do in other places](https://github.com/rack/rack-attack/blob/main/lib/rack/attack/store_proxy/dalli_proxy.rb#L10)?

I think this point is really important and we need to take the time to brainstorm/discuss it properly.

rack-attack.gemspec
johnnyshields: Please read my `spec/acceptance/error_handling_spec.rb`. I am using mocks to trigger various errors which do not normally occur and asserting how they are handled. `rspec-mocks` was designed exactly for this sort of use case (e.g. testing how novel errors propagate through a system.)

It is certainly possible to force certain errors by strange Redis configurations, this ultimately is limited in what types of errors we can trigger. I will be using mocks further in subsequent PRs.

lib/rack/attack.rb
santib: >In order to invoke rescue *allowed_errors, the allowed_errors variable must contain constants, not strings.

I know that. But why do we need to also allow `String`s? If supporting only constants makes the code simpler, then that works for me.

johnnyshields: As I've explained above, we cannot simply support only constants here. We must assume that `Rack::Attack` might be loaded before `Dalli` and `Redis`.

If we check for `defined?(Dalli)` at the time `Rack::Attack` loads, we may end up not setting those errors in the list. Unless we do something like dynamically converting strings to constants, which as I've illustrated above adds complexity.

If you would like to make a code suggestion on how this might be implemented, by all means please do. Otherwise, let's go forward with what I have here.

johnnyshields: one option would be to do:

```
def allowed_errors
  @allowed_errors || default_allowed_errors
end

def default_allowed_errors
   arr = []
   arr = Dalli::DalliError if defined?(Dalli)
   arr = Redis::BaseError if defined?(Redis)
   arr
end
```

I do not like this however, because:
1. doing `Rack::Attack.allowed_errors += [MyError]` will pin the value e.g. if Redis hasn't loaded yet
2. we are doing a dynamic build on every Rack::Attack call
3. the string approach is lower complexity/fewer LOC
4. as a bonus, the string approach also allows the user to define errors as strings for which consts may not have loaded yet

santib: I think there are other ways I said above

>I know you didn't like it before, but one way could be the Rack::Attack::CacheError class, but there could be others, for example having this class wrap and re-raise the original error, maybe? Or maybe having the defaults defined by each proxy class which also checks for defined?(Dalli) [as they already do in other places](https://github.com/rack/rack-attack/blob/main/lib/rack/attack/store_proxy/dalli_proxy.rb#L10)?

Some options:
1. `Rack::Attack::CacheError` as shown in https://github.com/rack/rack-attack/compare/main...santib:rack-attack:system-hardening-refactor#diff-9114718ab0fa2d856234c04d73b07c3758a6aaab1dbfbe87401f4eb4e358f127R134
2. `Rack::Attack::CacheError` as a wrapper for the original error, so we don't lose it
3. Having the store proxies dynamically add to an array of `allowed_errors` when they are invoked (to prevent the issue you are mentioning about load order)

I personally would prefer to go with 1. or 2., but I'd like to hear @grzuy thoughts on this.

johnnyshields: For reference, there is precedent in other Ruby libs to reference potentially unloaded classes with strings, such as Rails where you can see this:

```ruby
belongs_to :manager, class_name: "Employee"
```

johnnyshields: Last thing: this is a moot point. Please see these line in the intended final state of this series of PRs:

https://github.com/rack/rack-attack/pull/577/files#diff-9114718ab0fa2d856234c04d73b07c3758a6aaab1dbfbe87401f4eb4e358f127R218-R225

As you can see, I'm adding an additional error_handler here.

README.md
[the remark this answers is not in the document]
johnnyshields: Some reasons:
- If you have some complex logic inside your Rack::Attack handlers that raises errors.
- If you implemented your own store proxy (non-Redis/Dalli)

That being said I'd estimate 98%+ of users are fine with the default Redis/Dalli error handling.

lib/rack/attack.rb
johnnyshields: > Rack::Attack::CacheError

I would really prefer that we do not obscure the errors coming from Redis/Dalli itself. It makes it a lot harder to work with and understand the reasons why Redis/Dalli are failing.

johnnyshields:
@santib I feel these discussions are getting lost in the weeds. Ultimately this configurable error handling is not that important.

For now, I am fine to remove the `allowed_errors=` configuration and just define `allowed_errors` as follows:

```ruby
def allowed_errors
   arr = []
   arr << Dalli::DalliError if defined?(Dalli)
   arr << Redis::BaseError if defined?(Redis)
   arr
end
```

Would you accept this so we can move forward?

lib/rack/attack.rb
santib: >>Rack::Attack::CacheError
>
>I would really prefer that we do not obscure the errors coming from Redis/Dalli itself. It makes it harder to understand the reasons why Redis/Dalli are failing.

it could be just a wrapper, and then pass the original error to the application

johnnyshields: Let's defer this discussion until I get my other PRs merged.

johnnyshields:
Closing in favor of https://github.com/rack/rack-attack/pull/641. We'll revisit make `allowed_errors` configurable later.

PR #640, opened 2023-11-20
Freeze time to fix flakey specs

jdelStrother:
Noticed a CI failure here - https://github.com/rack/rack-attack/actions/runs/6932050109/job/18855154765 - which I think is due to the system clock rolling over 15:14:00 while mid-test. How about freezing time in these specs?

### DIFF
--- a/spec/rack_attack_throttle_spec.rb
+++ b/spec/rack_attack_throttle_spec.rb
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 require_relative 'spec_helper'
+require 'timecop'
 
 describe 'Rack::Attack.throttle' do
   before do
-    @period = 60 # Use a long period; failures due to cache key rotation less likely
+    @period = 60
     Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
     Rack::Attack.throttle('ip/sec', limit: 1, period: @period) { |req| req.ip }
   end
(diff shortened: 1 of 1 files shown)


jdelStrother:
Oh I missed that. Will send an update

spec/rack_attack_throttle_spec.rb
```diff
@@ -3,8 +3,9 @@
 require_relative 'spec_helper'
 
 describe 'Rack::Attack.throttle' do
+  freeze_time!
```
santib, line 6: I tried these changes but it didn't work
image

I think we need to make sure that `Timecop` gets called right before the request is sent (in another before block).

santib: also if it's possible to use `Timecop.freeze do` (with block) [I think I'd prefer it](https://github.com/travisjeffery/timecop#timecopsafe_mode) even if it means that we no longer have the requests performed in a before block.

jdelStrother: Oh, bah. Minitest::Spec doesn't support defining multiple `before` blocks at the same level 😞

jdelStrother: Thanks for catching that, I didn't test the timecop version well enough. See what you think to the new version.

santib (approved):

santib:
Thank you @jdelStrother 🙌

PR #641, opened 2023-11-20
System Hardening Phase 1: Catch errors at top level, take two

johnnyshields:
This PR removes the `rescuing` blocks from Dalli/RedisProxy classes and instead catches errors at the top level.

It is a simpler version of #639

### DIFF
--- a/lib/rack/attack.rb
+++ b/lib/rack/attack.rb
@@ -128,6 +128,17 @@ module Rack
         configuration.tracked?(request)
         @app.call(env)
       end
+    rescue *allowed_errors
+      @app.call(request.env)
+    end
+
+    private
+
+    def allowed_errors
+      errors = []
+      errors << Dalli::DalliError if defined?(Dalli)
+      errors << Redis::BaseError if defined?(Redis)
+      errors
     end
   end
 end
--- a/lib/rack/attack/store_proxy/dalli_proxy.rb
+++ b/lib/rack/attack/store_proxy/dalli_proxy.rb
@@ -24,34 +24,26 @@ module Rack
         end
 
         def read(key)
-          rescuing do
-            with do |client|
-              client.get(key)
-            end
+          with do |client|
+            client.get(key)
           end
         end
 
         def write(key, value, options = {})
-          rescuing do
-            with do |client|
-              client.set(key, value, options.fetch(:expires_in, 0), raw: true)
-            end
+          with do |client|
+            client.set(key, value, options.fetch(:expires_in, 0), raw: true)
           end
         end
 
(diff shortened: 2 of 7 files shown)


lib/rack/attack.rb
```diff
@@ -128,6 +128,17 @@ def call(env)
         configuration.tracked?(request)
         @app.call(env)
       end
+    rescue *allowed_errors
```
grzuy, line 131: On second thought, wouldn't this rescue dalli and redis errors ocurring inside other middlewares down the line called in `@app.call(env)`?

johnnyshields: You're right, I've refactored the scope which needs error handling into its own method.

johnnyshields:
@grzuy @santib Your help to review this PR is appreciated.

@santib in previous threads you had proposed introducing an intermediate error such as "StoreProxyError". I've expressed that I don't feel there is a big advantage in doing this, as it obscures the underlying errors which have valuable information. As you'll see in my [end-state](https://github.com/rack/rack-attack/pull/577) I'm adding a hook to report errors e.g. via Sentry/Bugsnag (user adds their own).

It would be great if we could merge this so I can proceed on the series of PRs that users are waiting on. If at the end there is a strong reason to introduce a StoreProxyError I'm happy to add it at the end.

santib:
Hey @johnnyshields,

I’d like to make clear that not necessarily all of the proposed PRs will get merged. We’ll need to evaluate the tradeoffs of each of them.

Regarding this PR, I think the end goal it’s solving is valuable and would like to see it merged once we agree on the solution :+1:

Regarding the solution, I still think the [wrapper error](https://github.com/rack/rack-attack/compare/main...santib:rack-attack:system-hardening-refactor#diff-9114718ab0fa2d856234c04d73b07c3758a6aaab1dbfbe87401f4eb4e358f127R134) is better ([this](https://github.com/rack/rack-attack/pull/641#discussion_r1401284751) wouldn’t have even happened), and maybe we should even consider going a step further and do something [like this](https://web.archive.org/web/20160303075815/http://devblog.avdi.org/2013/12/25/exception-causes-in-ruby-2-1/). The fact of raising a custom rack attack error doesn't mean we can't provide the original error to devs. But I’d like to hear @grzuy thoughts on this before moving on with either solution.

johnnyshields:
@santib lets keep it simple. What I have here is the lowest complexity implementation. We can always introduce more complexity at anytime in the future if it is warranted.

johnnyshields:
@santib @grzuy is there a way we can move forward here?

johnnyshields:
@santib can this be merged?

PR #642, opened 2023-12-13
ci: Update checkout action

santib:
Fix the following warning in the CI:

>The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/


grzuy (approved):
Nice.

We could consider automating this with dependabot `package-ecosystem: github-actions` https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem ?

PR #643, opened 2023-12-13
ci: freeze time in more specs

santib:
Example of failure https://github.com/santib/rack-attack/actions/runs/7199119786/job/19610109965

```
# Running:

........F

Failure:
ActiveSupport::Cache::MemoryStore as a cache backend#test_0002_works for fail2ban [/home/runner/work/rack-attack/rack-attack/spec/support/cache_store_helper.rb:36]:
Expected: 403
  Actual: 200

bin/rails test /home/runner/work/rack-attack/rack-attack/spec/support/cache_store_helper.rb:19

..........................................................................S....................................................................

Finished in 2.172762s, 69.9571 runs/s, 132.5502 assertions/s.
[15](https://github.com/santib/rack-attack/actions/runs/7199119786/job/19610109965#step:5:16)2 runs, 288 assertions, 1 failures, 0 errors, 1 skips
```

### DIFF
--- a/spec/support/cache_store_helper.rb
+++ b/spec/support/cache_store_helper.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require "timecop"
+require_relative 'freeze_time_helper'
 
 class Minitest::Spec
   def self.it_works_for_cache_backed_features(options)
@@ -11,7 +11,7 @@ class Minitest::Spec
         request.ip
       end
 
-      Timecop.freeze do
+      within_same_period do
         get "/", {}, "REMOTE_ADDR" => "1.2.3.4"
         assert_equal 200, last_response.status
 
@@ -27,7 +27,7 @@ class Minitest::Spec
         end
       end
 
-      Timecop.freeze do
+      within_same_period do
         get "/"
         assert_equal 200, last_response.status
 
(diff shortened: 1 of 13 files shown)


spec/support/cache_store_helper.rb
```diff
@@ -9,11 +11,13 @@ def self.it_works_for_cache_backed_features(options)
         request.ip
       end
 
-      get "/", {}, "REMOTE_ADDR" => "1.2.3.4"
-      assert_equal 200, last_response.status
+      Timecop.freeze do
```
grzuy, line 14: Thoughts on having a helper test method instead called something like `within_same_period` or something similar, to capture the spirit of the freezing?

grzuy (approved):

spec/support/cache_store_helper.rb
santib: there I made the changes. What do you think? I like it in some cases but in others where we freeze the time only to be able to generate the same key used by the cache it doesn't look so nice to me.

Also, any thoughts about dropping `Timecop` in favor of [ActiveSupport::Testing::TimeHelpers](https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html)?

grzuy: > Also, any thoughts about dropping `Timecop` in favor of [ActiveSupport::Testing::TimeHelpers](https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html)?

Not opposed to.

PR #649, opened 2024-01-10
Fix Rails test not being ran

santib:


spec/acceptance/cache_store_config_for_allow2ban_spec.rb
```diff
@@ -12,9 +12,11 @@
     end
   end
 
-  it "gives semantic error if no store was configured" do
-    assert_raises(Rack::Attack::MissingStoreError) do
-      get "/scarce-resource"
+  unless defined?(Rails)
```
santib, line 15: When using Rails the default `Rack::Attack.cache.store` is set to `Rails.cache`

santib:
Previously there were [152 runs](https://github.com/rack/rack-attack/actions/runs/7476598005/job/20347355061#step:5:19) for `test (3.2, rails_7_1)`

With these changes we are skipping 4 tests and adding 1 more that was incorrectly being skipped. And we see that there are [149 runs](https://github.com/rack/rack-attack/actions/runs/7476783862/job/20347995910#step:5:24) (152-4+1).

PR #654, opened 2024-03-04
Fix AS::Notification.subscribe arguments in README.md

tisba:
[ActiveSupport::Notifications.subscribe](https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html#method-c-subscribe)'s block arguments are as follows (taken from the Rails API docs):

```ruby
ActiveSupport::Notifications.subscribe('render') do |name, start, finish, id, payload|
  name    # => String, name of the event (such as 'render' from above)
  start   # => Time, when the instrumented block started execution
  finish  # => Time, when the instrumented block ended execution
  id      # => String, unique ID for the instrumenter that fired the event
  payload # => Hash, the payload
end
```

The README makes it look like the `id` argument would be the request ID, which is not the case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ Rack::Attack.track("special_agent", limit: 6, period: 60) do |req|
 end
 
 # Track it using ActiveSupport::Notification
-ActiveSupport::Notifications.subscribe("track.rack_attack") do |name, start, finish, event_id, payload|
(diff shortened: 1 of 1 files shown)


santib:
Good catch, thank you!

santib (approved):

PR #657, opened 2024-05-03
Refactor notifications checks in tests

santib:
- Unify how notifications are tracked and checked using `Array`'s `#push` `#pop` and `#size`.
- With this now we not only check "some notification happened" but the exact amount of notifications since in some cases we might trigger more than once, so this can prevent missing or mixing notifications.

Note: We could introduce a custom class built on top of Array with some facilities but seems to not be worth it at the time

### DIFF
--- a/spec/acceptance/blocking_ip_spec.rb
+++ b/spec/acceptance/blocking_ip_spec.rb
@@ -3,6 +3,8 @@
 require_relative "../spec_helper"
 
 describe "Blocking an IP" do
+  let(:notifications) { [] }
+
   before do
     Rack::Attack.blocklist_ip("1.2.3.4")
   end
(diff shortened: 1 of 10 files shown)


grzuy (approved):
Good call.
Neat.

Maybe adding some array emptying on after each test hook?
Just not to rely only on each test poping?
Minor thing, your call.

santib:
> Good call. Neat.
> 
> Maybe adding some array emptying on after each test hook? Just not to rely only on each test poping? Minor thing, your call.

Thanks for the review 😊 I think the emptying isn’t needed because the array gets recreated for each test. Or am I missing something?

grzuy:
> > Good call. Neat.
> > Maybe adding some array emptying on after each test hook? Just not to rely only on each test poping? Minor thing, your call.
> 
> Thanks for the review 😊 I think the emptying isn’t needed because the array gets recreated for each test. Or am I missing something?

Right, hehe.
My bad.

PR #681, opened 2024-12-28
Add Ruby 3.4 to the CI

santib:


gemfiles/active_support_6_redis_cache_store.gemfile
```diff
@@ -4,6 +4,9 @@ source "https://rubygems.org"
 
 gem "activesupport", "~> 6.1.0"
 gem "redis", "~> 5.0"
+gem "base64", "~> 0.2"
+gem "bigdecimal", "~> 3.1"
+gem "mutex_m", "~> 0.3"
```
grzuy, line 9: Why this exactly?
Needed by ruby 3.4 on these "old" activesupport versions?

santib: Ruby 3.4 extracted out some parts into their own gems, and no longer require them by default. Rails/Minitest depend on these, so even if it's fixed for newer versions, the older versions will fail. So we can either choose to drop support for Ruby 3.4 + old versions, or we can backfill those missing dependencies. Let me know if you come up with a better idea of course 😄 

https://github.com/rails/rails/pull/48907
https://github.com/minitest/minitest/pull/973

grzuy: Gotcha, thanks for the details!

So the fact that `activesupport` didn't backfill the dependency on those gems in activesupport < 7.1 for ruby 3.4, I read that implicitly as saying "activesupport < 7.1 isn't officially compatible with ruby 3.4".

I know that adding the deps to the Gemfile fixes it. But it seems maybe "too much coverage" on rack-attack side.

Feels like having ruby 3.4 tested against activesupport 7.1, 7.2, 8.0 and future upcoming versions seems fine to me.

Thoughts?

santib: I agree, sounds good to me. I'll exclude those combinations from the CI matrix

grzuy (approved):
:100:

PR #683, opened 2025-01-18
Fix CI

santib:
It started failing since `concurrent-ruby` released version `1.3.5`

https://github.com/rack/rack-attack/actions/runs/12424588943

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -2,6 +2,7 @@
 
 require "bundler/setup"
 
+require "logger"
 require "minitest/autorun"
 require "minitest/pride"
 require "rack/test"


rack-attack.gemspec
```diff
@@ -30,6 +30,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency 'appraisal', '~> 2.2'
   s.add_development_dependency "bundler", ">= 1.17", "< 3.0"
+  s.add_development_dependency 'concurrent-ruby', '<= 1.3.4'
```
santib, line 33: another option would be adding this to the spec gemfiles in `gemfiles/*` for the old versions of activesupport that fail, but we may need to drop support to ruby 2.5 and 2.6 that just fail completely.

That way we can still test concurrent-ruby 1.3.5+ for newer versions of ruby and rails

grzuy: Should we try fixing the tests by requiring the now missing logger instead of pinning to a legacy concurrent-ruby?
https://github.com/rails/rails/pull/54264.

santib: that's a great idea 😄 TY

grzuy (approved):
:clap:

PR #689, opened 2025-08-17
Bypasseable Errors for Rack Attack Proxy Instances

mpenalozag:
## Description
Some errors on proxy instances could lead to requests not being passed to the next middleware. This could be behaviour that we want to avoid, we wouldn't want our system to stop responding just because of unexpected issues between Rack::Attack and our external Cache instance.

## Considerations
- This is a short draft to validate the approach. Happy to iterate it.
- I noticed there could be more tests on some behaviours. Happy to add those after we agree on a solution


lib/rack/attack/store_proxy/dalli_proxy.rb
```diff
@@ -67,10 +67,10 @@ def with
           end
         end
 
-        def rescuing
-          yield
-        rescue Dalli::DalliError
-          nil
+        def should_bypass_error?(error)
+          # Dalli-specific default behavior: bypass Dalli errors
+          return true if defined?(::Dalli::DalliError) && error.is_a?(Dalli::DalliError)
```
santib, line 72: I like the idea to keep the existing behavior untouched, but maybe we can do something else to not have to override the `should_bypass_error?` method here?

maybe something like having this DalliProxy have `bypassable_store_errors` default to `['Dalli::DalliError']` as a String so that it doesn't fail if the gem is not installed?

mpenalozag: I think that's a good idea. Like you said, we could just define the `bypassable_store_errors` and let the `BaseProxy` class handle the rest.

lib/rack/attack/base_proxy.rb
```diff
@@ -5,6 +5,43 @@
 module Rack
   class Attack
     class BaseProxy < SimpleDelegator
+      attr_reader :bypass_all_store_errors, :bypassable_store_errors
+
+      def initialize(store, bypass_all_store_errors: false, bypassable_store_errors: [])
```
santib, line 10: wondering if it's better to have just one variable: `bypassable_store_errors` which can either by a `Symbol` (:all, :none), or an `Array`? Just to avoid the possible combinations of both variable sets:
- bypass_all_store_errors: false, bypassable_store_errors: [...]
- bypass_all_store_errors: true, bypassable_store_errors: []
- bypass_all_store_errors: false, bypassable_store_errors: [...]
- bypass_all_store_errors: true, bypassable_store_errors: []

mpenalozag: I think that's a good approach. We could also raise a configuration error if both variables are defined

santib:
Hey @mpenalozag, he may not be available at the time. So if you want we can go ahead with the PR 👍

lib/rack/attack/cache.rb
```diff
@@ -19,10 +19,10 @@ def initialize(store: self.class.default_store)
 
       attr_reader :store
 
-      def store=(store)
+      def store=(store, bypass_all_store_errors: false, bypassable_store_errors: [])
```
grzuy, line 22: How would this be called from the caller? 
Thinking about what the end user API would be.

grzuy (commented):
Hey there!

Sorry for the delay.

@mpenalozag thanks for you contribution!

Interested and eager to go a bit deeper on the issue/problem being addressed here and what others (if any) alternatives exist to address it, before considering adding a new feature that changes the public API in some way.

Can you elaborate on which particular proxy you're using and which errors in particular are making this a problem for you?

Thank you in advance!

mpenalozag:
@grzuy @santib Hey guys! Doing a follow up on this. Should I just add the tests and wait for you review?

mpenalozag:
@grzuy @santib Added tests and reordered commits for clarity

Issue #145, opened 2015-11-04
Behind a load balancer


klawrence:
I used the example throttle code

```
throttle('req/ip', limit: 100, period: 5.minutes) do |req|
  req.ip unless req.path.starts_with?('/assets')
end
```

This worked great on our staging server but immediately ran into the limit on production because req.ip returns the IP address of our load balancer and not the remote_ip of the client.

Note that remote_ip is a method in ActionDispatch::Request but not Rack::Attack::Request.

We are using Rails 3.2.2 on Ruby 2.2.2

I was able to workaround the problem with a new method on Rack::Attack::Request

```
class Rack::Attack
  class Request < ::Rack::Request
    def remote_ip
      @remote_ip ||= (env['action_dispatch.remote_ip'] || ip).to_s
    end
  end
end
```

and then

```
req.remote_ip unless req.path.starts_with?('/assets')
```

Maybe the README could suggest using remote_ip behind a load balancer?

ktheory:
First off, I totally sympathize with you about how weird it is that Rails' `remote_ip` and Rack's `ip` method can return different things based on how your reverse proxy is set up.

It's bonkers. :stuck_out_tongue_closed_eyes:

Thanks for noting it here where searchers can find it.

IMHO, this is a bug for Rails and/or Rack to fix so that `ActionDispatch::Request#remote_ip` returns the same thing as `Rack::Request#ip`. I'd heartily :+1: that pull request.

twk3:
@ktheory do you have a recommended approach for working around this? At the moment I'm planning to override rack request trusted_proxy? to use the same list I gave to action dispatch.

ryanmaclean:
@ktheory is it maybe advisable to make note of this distinction somewhere?

martyciz:
For Google App Engine env['action_dispatch.remote_ip'] still returned balancer IP, to get actual remote IP I added code like this:

```
def remote_ip
  remote = env['HTTP_X_FORWARDED_FOR'] ? env['HTTP_X_FORWARDED_FOR'].split(',')[0] : nil
  @remote_ip ||= (remote || ip).to_s
end
```

arturopie:
@martyciz if you are using that `remote_ip` for some kind of whitelist/secure thing, that code is not secure since the client could 'spoof' `HTTP_X_FORWARDED_FOR` and set it to what ever ip they want?
[...]

PR #698, opened 2025-10-16
Inherit Request class from ActionDispatch if available

bdewater-thatch:
I ran into https://github.com/rack/rack-attack/issues/145, but figured that inside a Rails app the least surprising thing would be for `req` to be like an `::ActionDispatch::Request` instead of patching my own `remote_ip` method in `Rack::Attack::Request`.

### DIFF
--- a/lib/rack/attack/request.rb
+++ b/lib/rack/attack/request.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-# Rack::Attack::Request is the same as ::Rack::Request by default.
+# Rack::Attack::Request is the same as :ActionDispatch::Request in Rails apps,
+# and ::Rack::Request in other apps by default.
 #
 # This is a safe place to add custom helper methods to the request object
 # through monkey patching:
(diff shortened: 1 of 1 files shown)


grzuy:
Hey @bdewater-thatch,

Thanks for the PR!

That's interesting, never though about that.

I think it you're right that would be the most unsurprising implementation if we were doing this for the first time.
Considering the amount of code out there calling methods on these req objects, we need to kind of "make sure" we're not breaking users rule blocks...

Need to take a deeper look to assess that I guess...

@santib thoughts?

bdewater-thatch:
Hey @grzuy, long time no see :)

I agree there's risk here of breaking changes but hopefully not too bad. The interfaces of both are very similar if not identical. 

My first sanity check was:
```rb
Rack::Request.instance_methods - ActionDispatch::Request.instance_methods
# => [] 
```

Then the sources itself:
- https://github.com/rack/rack/blob/v3.2.3/lib/rack/request.rb  includes `Env` and `Helpers` and implements only 4 instance methods. `delete_param` and `update_param` come from `Env but are overridden.
- https://github.com/rails/rails/blob/v8.0.3/actionpack/lib/action_dispatch/http/request.rb also includes `Rack::Request::Env` and `Rack::Request::Helpers`, implements `ip` directly, implements `params` via `ActionDispatch::Http::Parameters`, and gets `delete_param` and `update_param` from `Rack::Request::Env`.

